PR #4, opened 2025-04-24
simple implementation of list_providers

gautambaghel:


pkg/hashicorp/tfregistry/list.go
```diff
@@ -48,160 +24,11 @@ func ListOrganizations(getClient GetClientFn, t translations.TranslationHelperFu
...
-				// Handle API errors gracefully, maybe return as ToolResult error
-				// Check for specific TFE error types if needed
-				return mcp.NewToolResultError(fmt.Sprintf("failed to list organizations: %v", err)), nil
-			}
-
-			// Marshal the result (Items) to JSON
-			r, err := json.Marshal(result.Items)
-			if err != nil {
-				return nil, fmt.Errorf("failed to marshal organization list response: %w", err)
+			commonProviders := []string{
```
BBBmau, line 27: All looks good, just commenting here that this is static for testing purposes and will utilize the registry endpoint so that it's more dynamic.

gautambaghel: makes sense

PR #7, opened 2025-04-27
`tool`: add `listModules` (plural)

BBBmau:
i’ve got listModules and working on a single module, though it’s still WIP since i’m missing some details when only focusing on one modules, I also need to look into how to handle the names cause to access just one module name + namespace looks like this:
image

I also made a change to registry to use v1 though i believe for modules we use v1 and for registry provider information we use v2 so I'll need to add a condition for that.

### DIFF
--- a/e2e/e2e_test.go
+++ b/e2e/e2e_test.go
@@ -80,6 +80,33 @@ func TestE2E(t *testing.T) {
 		// TODO: this is static and should be updated to test with the actual API response.
 		require.Equal(t, "aws, google, azurerm, kubernetes, github, docker, null, random", textContent.Text, "expected providers to match")
 	})
+
+	// TODO: split the tests into multiple files
+	t.Run("CallTool list_modules", func(t *testing.T) {
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		defer cancel()
+
+		// When we call the "get_me" tool
+		request := mcp.CallToolRequest{}
+		request.Params.Name = "list_modules"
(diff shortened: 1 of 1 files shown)


cmd/hcp-terraform-mcp-server/gautam-mcp-test.log
gautambaghel, line 1: @BBBmau - can you pls remove this file?

pkg/hashicorp/types.go
```diff
@@ -0,0 +1,235 @@
...
+			PrevPage    any `json:"prev-page"`
+			TotalPages  int `json:"total-pages"`
+			TotalCount  int `json:"total-count"`
+		} `json:"pagination"`
+	} `json:"meta"`
+}
+
+// ProviderVersion represents the structure of the provider version response.
+// https://registry.terraform.io/v2/provider-versions/47756?include=provider-docs
+type ProviderVersionResponse struct {
```
gautambaghel: @BBBmau - you're not doing anything with these right? let me add them with my commits

gautambaghel: all the provider types

pkg/hashicorp/types.go
```diff
@@ -0,0 +1,235 @@
+package hashicorp
```
gautambaghel: this file can be inside tfregistry imo

BBBmau: this makes sense to do since these types are only tfregistry related types. good catch since i had intended to include it as part of the tfregistry directory

PR #10, opened 2025-04-30
Adding provider implementation for the MCP server

gautambaghel:
* ~~Needs fixes for e2e tests~~
* Needs fixes for pagination - will be added in a separate PR

### DIFF
--- a/e2e/e2e_test.go
+++ b/e2e/e2e_test.go
@@ -75,7 +75,7 @@ func TestE2E(t *testing.T) {
 		require.NoError(t, err, "expected to call 'providerDetails' tool successfully")
 
 		require.False(t, response.IsError, "expected result not to be an error")
-		require.Len(t, response.Content, 2, "expected content to have one item")
+		require.Len(t, response.Content, 1, "expected content to have one item")
 
 		textContent, ok := response.Content[0].(mcp.TextContent)
 		require.True(t, ok, "expected content to be of type TextContent")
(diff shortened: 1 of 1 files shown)


pkg/hashicorp/tfregistry/list.go
```diff
@@ -27,35 +29,124 @@ func ListProviders(registryClient *http.Client) (tool mcp.Tool, handler server.T
 			// pageNumber, _ := OptionalParam[int](request, "page_number")
 			// pageSize, _ := OptionalParam[int](request, "page_size")
 
-			commonProviders := []string{
-				"aws", "google", "azurerm", "kubernetes",
-				"github", "docker", "null", "random",
+			name := request.Params.Arguments["name"].(string)
```
BBBmau, line 32: are you still running into errors where the model attempts to send `{}` resulting in `interface{} nil error`?

pkg/hashicorp/tfregistry/utils.go
```diff
@@ -1,17 +1,195 @@
...
+			doc.Attributes.Path,
+		)
+	}
+
+	return response, nil
+}
+
+func GetProviderResourceDetails(client *http.Client, providerVersionID, resource string, logger *log.Logger) (string, error) {
+	var content string
+	for _, category := range []string{"resources", "data-sources"} {
```
BBBmau, line 139: is there a reason why we want to treat the `providerResourceDetails` to represent both `resources` and `data-sources`?

I ask this since even if we were to add pagination in a separate PR we would have to break this up by type. It paginates through the data-source first even if our desired type is a resource.

PR #17, opened 2025-05-01
Adding minor fixes

gautambaghel:


pkg/hashicorp/tfregistry/utils.go
```diff
@@ -132,8 +132,21 @@ func GetProviderResourceDetails(client *http.Client, version, name, namespace, s
 	for _, doc := range providerDocs.Docs {
 		// Include the doc if sourceType was not provided/empty OR if the doc category matches the provided sourceType
 		if !sourceTypeProvided || s == "" || doc.Category == s {
-			content += fmt.Sprintf("## %s \n\n**Id:** %s \n\n**Category:** %s\n\n**Subcategory:** %s\n\n**Path:** %s\n\n",
-				doc.Title, doc.ID, doc.Category, doc.Subcategory, doc.Path)
+			if match, err := containsSlug(sourceName.(string), doc.Slug); err == nil && match && doc.Language == "hcl" {
```
BBBmau, line 135: was it not working prior? Im not quite understanding at first glance why we need to perform another GET when we do so on line 119 that returns everything we need in one response.

BBBmau: nevermind this makes sense to do after going through the endpoint flow

PR #20, opened 2025-05-02
`tools`: split `listModules`; add `moduleDetails`

BBBmau:
improved UX by having modules match the same flow as providerDetails: below is an example

image

### DIFF
--- a/pkg/hashicorp/tfregistry/list.go
+++ b/pkg/hashicorp/tfregistry/list.go
@@ -113,6 +113,8 @@ var providerToNamespaceModule = map[string]interface{}{
 	"google":  []interface{}{"GoogleCloudPlatform", "terraform-google-modules"},
 	"aws":     []interface{}{"aws-ia", "terraform-aws-modules"},
 	"azurerm": []interface{}{"Azure", "aztfmod"},
+	"oracle":  []interface{}{"oracle", "oracle-terraform-modules"},
+	"alibaba": []interface{}{"alibaba", "terraform-alicloud-modules"},
 }
 
 func ListModules(registryClient *http.Client, logger *log.Logger) (tool mcp.Tool, handler server.ToolHandlerFunc) {


pkg/hashicorp/tfregistry/list.go
```diff
@@ -109,79 +109,121 @@ func providerResourceDetails(registryClient *http.Client, logger *log.Logger) (t
 
 const MODULE_BASE_PATH = "registry://modules"
 
+var providerToNamespaceModule = map[string]interface{}{
+	"google":  []interface{}{"GoogleCloudPlatform", "terraform-google-modules"},
+	"aws":     []interface{}{"aws-ia", "terraform-aws-modules"},
+	"azurerm": []interface{}{"Azure", "aztfmod"},
```
gautambaghel, line 115: can you add a few more here please?

"oracle": []interface{}{"oracle","oracle-terraform-modules"},
"alibaba": []interface{}{"alibaba","terraform-alicloud-modules"}

pkg/hashicorp/tfregistry/list.go
```diff
@@ -109,79 +109,121 @@ func providerResourceDetails(registryClient *http.Client, logger *log.Logger) (t
...
 
-	return listModulesTool, listModulesHandler
+	return moduleDetailsTool, moduleDetailsHandler
 }
 
-func getModuleDetails(providerClient *http.Client, namespace interface{}, name interface{}, currentOffset interface{}, logger *log.Logger) ([]byte, error) {
+func getModuleDetails(providerClient *http.Client, namespace interface{}, name interface{}, provider interface{}, currentOffset interface{}, logger *log.Logger) ([]byte, error) {
 	// Clean up the URI
 	uri := "modules"
 	if ns, ok := namespace.(string); ok && ns != "" {
```
gautambaghel, line 217: This takes care if namespace is not empty, what happens if the namespace is empty?

BBBmau: it continues with not modifying `uri`, the case of namespace being `nil` only happens when we're wanting to get all the modules - `registry.io/v1/modules`

pkg/hashicorp/tfregistry/list.go
BBBmau: done here: https://github.com/hashicorp/hcp-terraform-mcp-server/pull/20/commits/ac9c79737244a0d5b9cbb6ee06c9ccce923ba36c

PR #24, opened 2025-05-05
add repo license, codeowners and copywrite headers

mukeshjc:
release prep

### DIFF
--- a/.copywrite.hcl
+++ b/.copywrite.hcl
@@ -1,0 +1,14 @@
+schema_version = 1
+
+project {
+  license        = "MPL-2.0"
+  copyright_year = 2025
+
+  # (OPTIONAL) A list of globs that should not have copyright/license headers.
+  # Supports doublestar glob patterns for more flexibility in defining which
+  # files or folders should be ignored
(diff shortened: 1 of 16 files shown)


Dockerfile
```diff
@@ -1,3 +1,6 @@
+# Copyright (c) HashiCorp, Inc.
+# SPDX-License-Identifier: MPL-2.0
```
mukeshjc, line 2: We need not worry too much about the copyright headers for now. Once the repo goes public, we have [automation](https://github.com/hashicorp/copyright-notice-automation/?tab=readme-ov-file#automated-scans) built that will ensure all source files have this header and will auto-raise PRs for it.

.copywrite.hcl
```diff
@@ -0,0 +1,14 @@
+schema_version = 1
```
mukeshjc, line 1: This is the configuration file that our [internal copyright automation](https://github.com/hashicorp/copyright-notice-automation/?tab=readme-ov-file#automated-scans) depends on. Not a file that would need significant or frequent modifications.

PR #27, opened 2025-05-06
E2e tests for providerDetails and providerResourceDetails

gautambaghel:


pkg/hashicorp/tfregistry/handlers.go
```diff
@@ -21,36 +21,18 @@ func ProviderDetails(registryClient *http.Client, logger *log.Logger) (tool mcp.
...
-			// If the provider version doesn't exist, try the hashicorp namespace
-			if providerVersion == "" {
-				providerNamespace = "hashicorp"
-				providerVersion, err = GetLatestProviderVersion(registryClient, providerNamespace, providerName, logger)
-				if err != nil {
-					return nil, logAndReturnError(logger, fmt.Sprintf("Error getting latest provider version in %s namespace", providerNamespace), err)
-				}
+			// For typical provider and namespace hallucinations
+			defaultErrorGuide := "please check the provider name or the namespace, perhaps the provider is published under a different namespace or company name"
+			providerName, providerNamespace, providerVersion, providerDataType, err := resolveProviderDetails(request, registryClient, defaultErrorGuide, logger)
```
BBBmau: I like how you separated the logic so that it's in `utils.go`. We may want to consider a struct for provider info, having 4 outputs doesn't seem right.

Perhaps something like:
```go
type providerData struct {
  name string,
  namespace string,
  version string,
  type string
}
```

gautambaghel: yeah sounds good!

gautambaghel: Fixed!

Issue #25, opened 2025-05-05
add `v2` for `provider` related documentation


BBBmau:
we moved away from `v2` due to the `v1` endpoint returning all the details of a provider in one response, however we run into a case where the model will attempt to use `provider-reference` which is what we need if we want to provide information regarding `guides` and `ephemeral` values just to name a few.

A possible implementation (open to other solutions) is to use `v1` endpoint for all things resources/data-sources and only utilize `v2` when the mention of `provider-reference` is used. This will prevent multiple GET calls for resources/data-sources.

We found that the multiple GET requests for pagination caused poor UX on `v2`. Ideally we'd want to keep the `v1` implementation as we get all resources/data-sources in one single GET response.

List of endpoints we'd support under `v2`:
- `overview`: https://registry.terraform.io/v2/provider-docs?filter[provider-version]=70664&filter[category]=overview&filter[slug]=index&filter[language]=hcl
- `functions`: https://registry.terraform.io/v2/provider-docs?filter[provider-version]=70664&filter[category]=functions&filter[slug]=name_from_id&filter[language]=hcl
- `guides`: https://registry.terraform.io/v2/provider-docs?filter[provider-version]=70664&filter[category]=guides&filter[slug]=sql_instance_switchover&filter[language]=hcl

PR #33, opened 2025-05-09
Add v2 API support for ProviderDetails & providerResourceDetails

aditya2548:
Issue #25 
* Add v2 API support for ProviderDetails & providerResourceDetails for `providerDataType`: `functions`, `guides` and `overview`
* Reformat utils functions to reduce code duplication
* Add `sendPaginatedRegistryCall` function to make paginated registry calls for v2 API.
* Add test cases to support these changes

### DIFF
--- a/pkg/hashicorp/tfregistry/types.go
+++ b/pkg/hashicorp/tfregistry/types.go
@@ -333,20 +333,22 @@ type ProviderResourceDetails struct {
 // ProviderOverview represents the structure of the provider overview (how to use it) response.
 // https://registry.terraform.io/v2/provider-docs?filter[provider-version]=70800&filter[category]=overview&filter[slug]=index
 type ProviderOverview struct {
-	Data []struct {
-		Type       string `json:"type"`
-		ID         string `json:"id"`
-		Attributes struct {
-			Category    string `json:"category"`
-			Language    string `json:"language"`
-			Path        string `json:"path"`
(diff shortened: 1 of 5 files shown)


pkg/hashicorp/tfregistry/types.go
```diff
@@ -333,20 +333,22 @@ type ProviderResourceDetails struct {
...
-		} `json:"links"`
-	} `json:"data"`
+	Data []ProviderDocData `json:"data"`
+}
+
+type ProviderDocData struct {
+	Type       string `json:"type"`
+	ID         string `json:"id"`
+	Attributes struct {
+		Category    string      `json:"category"`
```
BBBmau, line 343: this is missing the content type that would provide us the most information from a body response especially when it comes to a guide:

image

Including content might be enough to improve the UX

aditya2548: Screenshot 2025-05-12 at 14 25 16
This was the response I received for this prompt. 

Regarding the content type field, this field is not returned in the `/v2/provider-docs?filter[provider-version]=70664&filter[category]=guides` API. `ProviderDetails` should be returning the list of functions/ guides/ resources/ data-sources, and then `ProviderResourceDetails` should be used to get the actual content of the function description/ guide (which is already implemented in this tool-> we parse the entire response of `/v2/provider-docs/{id}`)

PR #40, opened 2025-05-14
Fixed modules (moduleList & moduleDetails) and added test cases for it

gautambaghel:


e2e/payloads.go
```diff
@@ -208,4 +229,20 @@ var moduleDetailsTestCases = []RegistryTestCase{
...
+	{
+		TestShouldFail:  false,
+		TestDescription: "Testing listModules with Azure (aks) - no offset",
+		TestPayload: map[string]interface{}{
+			"moduleName":     "aks",
+			"moduleProvider": "azurerm",
+		},
+	},
+	{
+		TestShouldFail:  false,
```
BBBmau, line 241: we should include a test case where we expect a failure since we still run into an expected failure when no module is found

image

Issue #43, opened 2025-05-14
add support for tool annotations


aditya2548:
We should add support for tool annotations to the existing server tools. These annotations serve as metadata to help describe each tool, improving the overall user experience by providing better context and guidance.

For more details, refer to the documentation: [Tool Annotations – Model Context Protocol](https://modelcontextprotocol.io/docs/concepts/tools#tool-annotations)

gautambaghel:
Merged with #44

PR #44, opened 2025-05-14
feature: add tool annotations

aditya2548:
Addresses issue #43 by adding tool annotations for existing tfregistry tools. This will help improve the user experience as clients would have better understanding of the tools.

### DIFF
--- a/pkg/hashicorp/tfregistry/handlers.go
+++ b/pkg/hashicorp/tfregistry/handlers.go
@@ -23,6 +23,8 @@ func ResolveProviderDocID(registryClient *http.Client, logger *log.Logger) (tool
 			Use the most relevant single word as the search query for serviceSlug, if unsure about the serviceSlug, use the providerName for its value.
 			When selecting the best match, consider: - Title similarity to the query - Category relevance Return the selected providerDocID and explain your choice.  
 			If there are multiple good matches, mention this but proceed with the most relevant one.`),
+			mcp.WithTitleAnnotation("Identify the most relevant provider document ID for a Terraform service"),
+			mcp.WithReadOnlyHintAnnotation(true),
 			mcp.WithString("providerName", mcp.Required(), mcp.Description("The name of the Terraform provider to perform the read or deployment operation")),
 			mcp.WithString("providerNamespace", mcp.Required(), mcp.Description("The publisher of the Terraform provider, typically the name of the company, or their GitHub organization name that created the provider")),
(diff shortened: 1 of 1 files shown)


pkg/hashicorp/tfregistry/handlers.go
```diff
@@ -93,6 +96,9 @@ func providerResourceDetails(registryClient *http.Client, logger *log.Logger) (t
 	return mcp.NewTool("providerResourceDetails",
 			mcp.WithDescription(`This tool is used to obtain the documentation, schema, and code examples from a given Terraform provider version, which will guide you in deploying a specific service on cloud, on-premise, and SaaS application environments. 
 			Please specify the provider name, namespace, and the service name you wish to provision to utilize this tool.`),
+			mcp.WithTitleAnnotation("Provider Resource Details"),
+			mcp.WithReadOnlyHintAnnotation(true),
+			mcp.WithOpenWorldHintAnnotation(true),
```
BBBmau, line 116: we can remove this since by default they are set to true already for `openWorldHintAnnotation`. It makes sense to include ReadOnlyHint since by default it's false.

aditya2548: Makes sense, Done!

gautambaghel: what's the default value for `WithIdempotentHintAnnotation` we should add that as true also

gautambaghel: I just looked, it's false -https://modelcontextprotocol.io/docs/concepts/tools#available-tool-annotations

we should explicitly set it to true for all the tools

BBBmau: @gautambaghel  it's false by default but it seems like it only makes sense to set to true if readOnly is false, from docs:
image

gautambaghel: interesting, okay we can leave it as it is

pkg/hashicorp/tfregistry/handlers.go
```diff
@@ -23,6 +23,8 @@ func ProviderDetails(registryClient *http.Client, logger *log.Logger) (tool mcp.
 			It helps users understand everything that can be provisioned and managed using the Terraform provider by listing out its resources (write operations), data sources (read operations), and functions (utility operations). 
 			For each item, note the existence and path of its documentation.
 			`),
+			mcp.WithTitleAnnotation("Provider Details"),
```
gautambaghel: It seems title is used for human reading, can we explain what this tool does in 4-5 words?

aditya2548: Updated the annotations as per new handlers to be more descriptive

gautambaghel (approved):
LGTM!

PR #45, opened 2025-05-14
`feature`: replace `listModules` -> `searchModules`

BBBmau:


pkg/hashicorp/tfregistry/handlers.go
```diff
@@ -132,66 +130,44 @@ func providerResourceDetails(registryClient *http.Client, logger *log.Logger) (t
...
-	return mcp.NewTool("listModules",
-			mcp.WithDescription(`This tool helps users deploy complex services on cloud and on-premise environments by retrieving a list of Terraform modules.
-			Please specify the provider name to utilize this tool. You can also use this tool without specifying a provider to get a list of all available modules.`),
-			mcp.WithString("moduleProvider",
-				mcp.Description("The name of the provider for the Terraform module to use."),
+func SearchModules(registryClient *http.Client, logger *log.Logger) (tool mcp.Tool, handler server.ToolHandlerFunc) {
+	return mcp.NewTool("searchModules",
+			mcp.WithDescription(`This tool helps users deploy complex services on cloud and on-premise environments by searching for a list of Terraform modules.
+			Please specify the provider name to utilize this tool. You can also use this tool without specifying a provider to get a list of all available modules. you are REQUIRED to call moduleDetails tool after searchModules to get the moduleID. If no modules were found, reattempt the search with a new moduleName query.`),
```
gautambaghel: @BBBmau - we don't need providerName anymore, right? There's no place to input it

Can we use the description something like this


This tool helps users deploy complex services on cloud and on-premise environments by searching for a list of Terraform modules. It resolves a module name to obtain a compatible moduleID for the moduleDetails tool and returns a list of matching libraries. You MUST call this function before 'moduleDetails' to obtain a valid and compatible moduleID. When selecting the best match, consider: - Name similarity to the query - Description relevance - Code Snippet count (documentation coverage) - Download counts (popularity) Return the selected moduleID and explain your choice. If there are multiple good matches, mention this but proceed with the most relevant one. If no modules were found, reattempt the search with a new moduleName query.

pkg/hashicorp/tfregistry/handlers.go
```diff
@@ -132,66 +130,44 @@ func providerResourceDetails(registryClient *http.Client, logger *log.Logger) (t
...
-				content, err := UnmarshalTFModulePlural(response)
-				if err != nil {
-					return nil, logAndReturnError(logger, "unmarshalling modules", err)
-				}
-				return mcp.NewToolResultText(*content), nil
-			}
-
-			if mp, ok := moduleProvider.(string); !ok {
-				return nil, logAndReturnError(logger, "error finding the provider, It represents the provider associated with the module, typically the name of the provider where most resources are deployed like aws, azurerm, google etc.", nil)
+			if mq, ok := moduleQuery.(string); !ok && moduleQuery == nil {
```
gautambaghel: I don't think you need `moduleQuery == nil` here, do you? can you point me to the test case where you're testing this?

pkg/hashicorp/tfregistry/handlers.go
```diff
@@ -132,66 +130,44 @@ func providerResourceDetails(registryClient *http.Client, logger *log.Logger) (t
...
-					logger.Infof("Trying to use %v as potential sources of modules for the provider %s", potentialModuleNamespaces, mp)
-				}
-
 				var modulesData, errMsg string
-				for _, moduleNamespace := range potentialModuleNamespaces {
-					response, err := GetModuleDetails(registryClient, ModuleDetail{ModuleNamespace: moduleNamespace}, currentOffsetValue, logger)
+				response, err := searchModules(registryClient, mq, currentOffsetValue, logger)
+				if err != nil {
+					errMsg = fmt.Sprintf("no module(s) found for moduleName: %s, error: %s", mq, err)
+					return nil, logAndReturnError(logger, errMsg, nil)
```
gautambaghel: combine these 2 lines

`return nil, logAndReturnError(logger, fmt.Sprintf("no module(s) found for moduleName: %s, mq), err)`

pkg/hashicorp/tfregistry/handlers.go
```diff
@@ -202,58 +178,32 @@ func ListModules(registryClient *http.Client, logger *log.Logger) (tool mcp.Tool
...
-				mcp.Required(),
-				mcp.Description("The name of the module to to access its detailed information."),
-			),
-			// TODO: We shouldn't need to include provider as an input, we could potentially grab the provider value from first GET and then perform a second GET with the provider value
-			mcp.WithString("moduleProvider",
+		To use it, please specify the module name and its associated provider. It's required that searchModules is used first to get the moduleID.`),
+			mcp.WithString("moduleID",
 				mcp.Required(),
-				mcp.Description("The provider associated with the module, it's the name of the provider where most resources are deployed like aws, azurerm, google etc. used to determine the correct namespace or the publisher of the module."),
+				mcp.Description("The ID of the module to to access its detailed information."),
```
gautambaghel: Here's a better description

Exact valid and compatible moduleID retrieved from searchModules (e.g., 'squareops/terraform-kubernetes-mongodb/mongodb/2.1.1', 'GoogleCloudPlatform/vertex-ai/google/0.2.0')

pkg/hashicorp/tfregistry/handlers.go
```diff
@@ -202,58 +178,32 @@ func ListModules(registryClient *http.Client, logger *log.Logger) (tool mcp.Tool
...
 	return mcp.NewTool("moduleDetails",
 			mcp.WithDescription(`This tool provides comprehensive details about a Terraform module, including inputs, outputs, and examples, enabling users to understand its effective usage. 
-		To use it, please specify the module name and its associated provider.`),
-			mcp.WithString("moduleName",
-				mcp.Required(),
-				mcp.Description("The name of the module to to access its detailed information."),
-			),
-			// TODO: We shouldn't need to include provider as an input, we could potentially grab the provider value from first GET and then perform a second GET with the provider value
-			mcp.WithString("moduleProvider",
+		To use it, please specify the module name and its associated provider. It's required that searchModules is used first to get the moduleID.`),
```
gautambaghel: Fetches up-to-date documentation on how to use a Terraform module. You must call 'searchModules' first to obtain the exact valid and compatible moduleID required to use this tool.

gautambaghel: ^^ better description

e2e/payloads.go
```diff
@@ -148,148 +148,110 @@ var providerDetailsTestCases = []RegistryTestCase{
...
+			"moduleQuery": "unknownprovider",
 		},
 	},
 	{
 		TestShouldFail:  false,
-		TestDescription: "Testing listModules with vSphere (capitalized)",
+		TestDescription: "Testing searchModules with vSphere (capitalized)",
 		TestPayload: map[string]interface{}{
-			"moduleProvider": "vSphere",
+			"moduleQuery": "vSphere",
```
gautambaghel, line 204: Can you add examples with spaces and a non existent moduleQuery

pkg/hashicorp/tfregistry/handlers.go
BBBmau: just tested without it and it isn't necessary. good catch!

e2e/payloads.go
BBBmau: add example with spaces, we have examples of non existent moduleQuery [here](https://github.com/hashicorp/terraform-mcp-server/blob/b38a1996e254a13bc86e7dfd7523a2db20d69a92/e2e/payloads.go#L152-L161)

Issue #48, opened 2025-05-15
bug: currentOffset value is not getting parsed in SearchModules


aditya2548:
currentOffset value is not getting parsed in SearchModules tool, and we're defaulting to 0 value. This hinders client from getting all the modules data.

PR #49, opened 2025-05-15
fix: currentOffset parsing for searchModules

aditya2548:
This PR fixes issue #48 as `number` is deserialized as `float64`

### DIFF
--- a/pkg/hashicorp/tfregistry/handlers.go
+++ b/pkg/hashicorp/tfregistry/handlers.go
@@ -7,8 +7,8 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"net/http"
 	log "github.com/sirupsen/logrus"
+	"net/http"
 
 	"github.com/mark3labs/mcp-go/mcp"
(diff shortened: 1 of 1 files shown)


pkg/hashicorp/tfregistry/handlers.go
```diff
@@ -146,8 +146,8 @@ func SearchModules(registryClient *http.Client, logger *log.Logger) (tool mcp.To
 			moduleQuery := request.Params.Arguments["moduleQuery"]
 			currentOffset := request.Params.Arguments["currentOffset"]
 			currentOffsetValue := 0
-			if _, ok := currentOffset.(int); ok {
-				currentOffsetValue = currentOffset.(int)
+			if val, ok := currentOffset.(float64); ok {
+				currentOffsetValue = int(val)
```
gautambaghel, line 150: So wouldn't this fail? If the value is float64 and downcasted to int?

aditya2548: It'll truncate the decimal portion of the float64. Also, the api accepts int values as acceptable offset

PR #50, opened 2025-05-15
fix: provider tools

gautambaghel:


pkg/hashicorp/tfregistry/handlers.go
```diff
@@ -7,48 +7,37 @@ import (
...
+	return mcp.NewTool("resolveProviderDocID",
+			mcp.WithDescription(`This tool retrieves a specific Terraform provider version. You MUST call this function before 'getProviderDocs' to obtain a valid tfprovider-compatible providerDocID. 
+			When selecting the best match, consider: - Name similarity to the query - Description relevance Return the selected providerDocID and explain your choice. 
+			If there are multiple good matches, mention this but proceed with the most relevant one.`),
+			mcp.WithString("providerName", mcp.Required(), mcp.Description("The name of the Terraform provider to perform the read or deployment operation")),
+			mcp.WithString("providerNamespace", mcp.Required(), mcp.Description("The publisher of the Terraform provider, typically the name of the company, or their GitHub organization name that created the provider")),
+			mcp.WithString("serviceName", mcp.Required(), mcp.Description("The name of the service you want to deploy or read using the Terraform provider")),
+			mcp.WithString("providerVersion", mcp.Description("The version of the Terraform provider to retrieve in the format 'x.y.z', or 'latest' to get the latest version")),
```
BBBmau, line 33: are we not missing a category section here? I'm attempting to access a guide to setup external credentials but getting no results:

image

gautambaghel: The error message is wrong and unhelpful, too

gautambaghel: @BBBmau - we don't need a category, Mauricio. What we do is we output a category for each of the document types, and based on the serviceName match and context given by the end user, we let AI pick the best category, which happens in the `getProviderDocs` tool

BBBmau: here's another reason to bring back a category input for the model, it shows all docs even though guides is the only thing desired. it uses "google" as the `serviceName`.

image

Issue #37, opened 2025-05-12
public repository preparation and addition of release workflows


mukeshjc:
- from doormat side we would need a proposal to change repo visibility : https://next.doormat.hashicorp.services/catalog/github/github-repo/35b10080e2
  - I'm unsure if security will vet this proposal or just repo owners can approve it.

- then for the security checlist - most items are completed
https://docs.google.com/spreadsheets/d/1DrKdK_3Qr6jGQqND-KESASDvSqeFbkReMKL3Dku7m1E/edit?gid=1359583161#gid=1359583161
  - few items left at the end of sheet. also for the follow-ups marked Yellow, I have received some responses from folks, so going to add them to the doc.

mukeshjc:
Resolving this issue with the initial rollout announced : https://developer.hashicorp.com/terraform/docs/tools/mcp-server

PR #52, opened 2025-05-15
[release-prep] CRT onboarding for release management

mukeshjc:
Complete CRT release management configuration and build steps.

1. Sample execution of build workflow ([link](https://github.com/hashicorp/terraform-mcp-server/actions/runs/15059429738)). 
   - executes on push for every branch in the repository to ensure all builds work for the changes on feature branches
   - we support binary releases for most os/aarch platforms. docker images are supported for linux systems using `gcr.io/distroless/base-debian12` as base image.
2. Developer UX remains the same. Makefile additions are only focused on adding `crt-build` target. 
   - `make build`, `make docker-build` are the targets developers continue to use for local development.
   - `make build` now understands the underlying os/aarch of the system to build the binary.
3. Security scanner configuration is added for ([PSA-1952](https://hashicorp.atlassian.net/browse/PSA-1952)) fix.
4. Version string is centralized into `version/VERSION` file.
   - used in `.github/workflows/build.yml` file for tagging the binaries and docker images
   - used in `version/version.go` so that `./terraform-mcp-server -v` can work reliably
   
   - > image

[PSA-1952]: https://hashicorp.atlassian.net/browse/PSA-1952?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

#37


.github/workflows/build.yml
```diff
@@ -0,0 +1,173 @@
...
+# We now default to running this workflow on every push to every branch.
+# This provides fast feedback when build issues occur, so they can be
+# fixed prior to being merged to the main branch.
+#
+# If you want to opt out of this, and only run the build on certain branches
+# please refer to the documentation on branch filtering here:
+#
+#   https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushbranchestagsbranches-ignoretags-ignore
+#
+on: [workflow_dispatch, push]
```
gautambaghel, line 12: Is it gonna create releases for every push in every branch? Wouldn't that be too much?

.github/workflows/build.yml
```diff
@@ -0,0 +1,173 @@
...
+    name: Docker ${{ matrix.arch }} build
+    needs:
+      - set-product-version
+      - build
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+        #   - { arch: "386" }
+        #   - { arch: "arm" }
```
gautambaghel, line 122: might as well? any particular reason we don't wanna do 386 and arm?

README.md
```diff
@@ -124,10 +124,11 @@ The following sets of tools are available:
 
 | Toolset     | Tool                       | Description                                                                                                           |
 |-------------|----------------------------|-----------------------------------------------------------------------------------------------------------------------|
-| `providers` | `providerDetails`          | Get comprehensive information about a specific provider, including its resources, data sources, and functions         |
+| `providers` | `providerDetails`          | Get comprehensive information about a specific provider, including its resources, data sources, functions, guides, and overview |
```
gautambaghel, line 127: This is going to change after #50 merges

.github/workflows/build.yml
mukeshjc: It won't create releases (into artifactory for example) but only run the workflow on every push to all branches. These executions will eventually reflect in the "Github Checks" section, when a PR will be submitted for that branch.

Also, this is the suggested configuration from CRT documentation.

.github/workflows/build.yml
mukeshjc: The docker image we are using as base doesn't support these platforms specifically in their manifest files. Hence I ignored those. We can change the base image to `alpine` for example and be able to support all these platforms, while only increasing the image size a bit compared to the distroless images which are most compact in size. 

I propose we use `alpine` as base, and open to make that change in a follow-up PR. Thoughts?

---
```
ERROR: failed to solve: gcr.io/distroless/base-debian12: 
  failed to resolve source metadata for gcr.io/distroless/base-debian12:latest: 
    no match for platform in manifest: not found
```


arm : https://github.com/hashicorp/terraform-mcp-server/actions/runs/15056253232/job/42322819816#step:3:344

386 : https://github.com/hashicorp/terraform-mcp-server/actions/runs/15056056687/job/42322235840#step:3:344

README.md
mukeshjc: Yes. My PR is only making the README consistent with the latest changes on main.

PR #53, opened 2025-05-16
`tests`: add unit_tests for `utils.go` #1

BBBmau:
to prevent a massive PR, a second PR will be made that will introduce unit tests for util functions used by provider tools. 

Includes a GHA for running unit tests that covers the following:
- sendRegistryCall
- UnMarshalling
- ProvderNameVersionExtraction
- SlugCheck
- LogReturn
- ProviderDataTypeCheck


.github/workflows/unit_test.yaml
```diff
@@ -0,0 +1,27 @@
+name: Unit Tests
```
aditya2548, line 4: Can we please add the Copyright header and newline for this file?

pkg/hashicorp/tfregistry/utils.go
```diff
@@ -541,10 +541,10 @@ func sendPaginatedRegistryCall[T any](client *http.Client, uriPrefix string, log
...
-	if err == nil {
-		err = fmt.Errorf("%s", context)
-	}
-	logger.Errorf("Error in %s: %v", context, err)
+    err = fmt.Errorf("%s, %w", context, err)
+    if logger != nil {
+        logger.Errorf("Error in %s, %v", context, err)
+    }
 	return err
 }
```
aditya2548: Can we change this to handle nil errors properly (having a nil error logs `%!w(<nil>)`), and to avoid logging context twice? Something like:

```suggestion
func logAndReturnError(logger *log.Logger, context string, err error) error {
	if err != nil {
		err = fmt.Errorf("%s: %w", context, err)
	} else {
		err = fmt.Errorf("%s", context)
	}
	if logger != nil {
        logger.Errorf("Error in %v", err)
	}
	return err
}
```

pkg/hashicorp/tfregistry/utils_test.go
```diff
@@ -0,0 +1,290 @@
+// +build !integration
```
aditya2548: Can we please add Copyright header for this file as well?

pkg/hashicorp/tfregistry/utils_test.go
aditya2548, line 1: Can we please combine multiple related test functions into one. For example:
```
func TestSendRegistryCall(t *testing.T) {
	tests := []struct {
		name      string
		uri       string
		version   string
		expectErr string
	}{
		{"Success_v1", "providers/hashicorp/aws", "v1", ""},
		{"Success_v2", "provider-docs?filter[provider-version]=6221", "v2", ""},
		{"404NotFound_v1", "test-uri", "v1", "error: 404 Not Found"},
		{"404NotFound_v2", "test-uri", "v2", "error: 404 Not Found"},
	}

	for _, tc := range tests {
		t.Run(tc.name, func(t *testing.T) {
			_, err := sendRegistryCall(client, "GET", tc.uri, logger, "v2")

			if tc.expectErr == "" && err != nil {
				t.Fatalf("expected no error, got %v", err)
			}
			if tc.expectErr != "" && (err == nil || err.Error() != tc.expectErr) {
				t.Errorf("expected %q, got %v", tc.expectErr, err)
			}
		})
	}
}

```

pkg/hashicorp/tfregistry/utils_test.go
```diff
@@ -0,0 +1,290 @@
...
+
+// --- sendRegistryCall ---
+
+var client = &http.Client{
+	Transport: &http.Transport{
+		Proxy: http.ProxyFromEnvironment,
+	},
+}
+
+var logger = log.New()
```
aditya2548: Instead of global client and logger, can we have these isolated in each test itself?

pkg/hashicorp/tfregistry/utils_test.go
aditya2548, line 1: Can we please use `stretchr/testify` for assertions in this test suite using require/ assert

pkg/hashicorp/tfregistry/utils_test.go
```diff
@@ -0,0 +1,290 @@
...
+
+func TestSendRegistryCall_Success_v1(t *testing.T) {
```
aditya2548: Can we use `httptest.Server` to mock HTTP responses for deterministic and isolated tests instead of relying on real HTTP calls?

aditya2548 (changes requested):
Thanks for this PR! Have added some suggestions

pkg/hashicorp/tfregistry/utils_test.go
BBBmau: this makes sense to do, thanks!

BBBmau: this makes sense, this would considerably simplify our test suite if we use `testify`. I'll add this in the next commits

pkg/hashicorp/tfregistry/utils_test.go
BBBmau: this was addressed with the use of httptest.Server implementation

pkg/hashicorp/tfregistry/utils_test.go
BBBmau: just implemented

pkg/hashicorp/tfregistry/utils.go
aditya2548: Can we please check this as well

BBBmau:
> @BBBmau Just one comment which I think we might have missed previously. Rest all looks great! Thanks for this

i think this is good as is, there's no need to do checks on whether err != nil or not due to the function itself only being called when err != nil.

gautambaghel (approved):
Approving, LGTM!

PR #56, opened 2025-05-19
[release-prep] update base docker image to alpine for supporting 386 and arm (32 bit) arch types

mukeshjc:
`docker.mirror.hashicorp.services/alpine:3.21` is a common choice for our OSS product images. [example consul](https://github.com/hashicorp/consul/blob/main/Dockerfile#L16C1-L19)

Previous distroless Docker image didn't support 386 and arm platforms.

image

#37

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,3 +1,6 @@
+# Copyright (c) HashiCorp, Inc.
+# SPDX-License-Identifier: MPL-2.0
+
 name: build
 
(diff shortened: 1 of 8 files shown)


.github/workflows/build.yml
aditya2548, line 1: Can we please add Copyright headers to workflow yml files (build and e2e_test) as well?

mukeshjc: 👍🏻  Done.

PR #59, opened 2025-05-19
[release] remove pre-promotion and promotion events

mukeshjc:
- enos is a test tool that should be configured if we use it only
- we don't have an iron-bank repo and nor did we need auto-version-bump for release


aditya2548 (dismissed):
LGTM!

PR #60, opened 2025-05-19
add security scanner, bump go & review GHA workflows

dduzgun-security:


.github/workflows/security-scan.yml
```diff
@@ -0,0 +1,75 @@
...
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          repository: hashicorp/security-scanner
+          token: ${{ secrets.PRODSEC_SCANNER_READ_ONLY }}
+          path: security-scanner
+          ref: main
+
+      - name: Scan
+        id: scan
+        uses: ./security-scanner
```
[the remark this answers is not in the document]
dduzgun-security: We purposefully don't do this.

.github/workflows/security-scan.yml
```diff
@@ -0,0 +1,75 @@
...
+    # The first check ensures this doesn't run on community-contributed PRs, who
+    # won't have the permissions to run this job.
+    if: ${{ (github.repository != 'hashicorp/terraform-mcp-server' || (github.event.pull_request.head.repo.full_name == github.event.pull_request.base.repo.full_name))
+      && (github.actor != 'dependabot[bot]') && (github.actor != 'team-proj-mcp-servers') }}
+
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5 # v5.5.0
+        with:
+          cache: ${{ contains(runner.name, 'Github Actions') }}
```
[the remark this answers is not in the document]
dduzgun-security: N/A

BBBmau (approved):
looks good to me, thanks for working on this!

.github/dependabot.yml
```diff
@@ -0,0 +1,13 @@
+# Copyright (c) HashiCorp, Inc.
+# SPDX-License-Identifier: BUSL-1.1
```
mukeshjc: **Issue:** this repository is MPL-2.0 , so that copyright would apply to this workflow file also right?

.github/workflows/build.yml
```diff
@@ -12,15 +12,23 @@ name: build
...
   PKG_NAME: "terraform-mcp-server"
 
+permissions:
+  contents: read
+
 jobs:
   get-go-version:
     name: "Determine Go toolchain version"
-    runs-on: ubuntu-latest
+    runs-on: ${{ github.repository == 'hashicorp/terraform-mcp-server' && 'ubuntu-latest' || fromJSON('["self-hosted","ondemand","os=linux"]') }}
```
mukeshjc: **Question:** Is there a known case where the `self-hosted` linux box would be used?

I believe - this workflow is a not a reusable one (there is no `workflow_call` property for it) , so it wouldn't be triggered by any other repository apart from the current one, hence the `github.repository` property would always be equal to `hashicorp/terraform-mcp-server` meaning we only use `ubuntu-latest`.

scan.hcl
```diff
@@ -0,0 +1,69 @@
+# Copyright (c) HashiCorp, Inc.
+# SPDX-License-Identifier: BUSL-1.1
```
mukeshjc: **Issue:** copyright license issue, repo uses MPL-2.0

.github/workflows/e2e_test.yaml
```diff
@@ -12,7 +12,7 @@ permissions:
 
 jobs:
   e2e-test:
-    runs-on: ubuntu-latest
+    runs-on: ${{ github.repository == 'hashicorp/terraform-mcp-server' && 'ubuntu-latest' || fromJSON('["self-hosted","ondemand","os=linux"]') }}
```
mukeshjc: **Question:** same as build.yml file, this workflow is not reusable one and hence wouldn't be run outside of current repository?

.github/dependabot.yml
gautambaghel: Yes, for reference: https://github.com/hashicorp/terraform-provider-boundary/blob/main/.github/dependabot.yml#L2

gautambaghel: @dduzgun-security - mind making this small update?

.github/workflows/build.yml
dduzgun-security: You are right for the `workflow_call` and this would actually never trigger that inside of the self-hosted runner. I'll go ahead and edit this.

.github/dependabot.yml
dduzgun-security: Should be resolved, thanks for catching this

scan.hcl
dduzgun-security: Should be resolved, thanks for catching this

mukeshjc (approved):

PR #62, opened 2025-05-20
add CRT fixes made for 0.1.0 release into `main` for future releases

mukeshjc:
These fixes were made as part of 0.1.0 release and have to be added back to `main` so future releases are smoother

### DIFF
--- a/.release/ci.hcl
+++ b/.release/ci.hcl
@@ -66,14 +66,6 @@ event "promote-staging" {
     on = "always"
   }
 
-  promotion-events {
-
-    pre-promotion {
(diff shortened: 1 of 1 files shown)


mukeshjc:
> Should this be running the promotion steps?

@aditya2548 
Yes, we will have CRT steps running on `main` branch as it produces Dev artifacts that we can use for testing. And as per our release 0.1.0 learning, these promotion events (deleted in the PR) are additional to the actual promotion of staging and release artifacts, which are not applicable to your product.

aditya2548 (approved):

PR #64, opened 2025-05-20
feature: Add ListOrganizations & GetOrganizationDetails tfe tools

aditya2548:
### Changes
* Introduced the initial package structure for Enterprise Tools.
* Added support for the `HCP_TFE_TOKEN` and `HCP_TFE_ADDRESS` environment variables.
* Implemented the following Terraform Enterprise organization tools:
  * ListOrganizations
  * GetOrganizationDetails
* Note: Updates to end-to-end (e2e) tests to reflect these changes will be included in a separate PR.
* Clients can invoke these tools sequentially to retrieve relevant organization data. For example:
Screenshot 2025-05-19 at 15 57 38


pkg/hashicorp/tfenterprise/organization/handlers.go
```diff
@@ -0,0 +1,89 @@
...
+func ListOrganizations(tfeClient *tfe.Client, logger *log.Logger) (tool mcp.Tool, handler server.ToolHandlerFunc) {
+	return mcp.NewTool("listOrganizations",
+			mcp.WithDescription("This tool lists all organizations the authenticated user has access to in Terraform cloud/ enterprise. An optional query can be passed, which helps match organizations by name or email."),
+			mcp.WithTitleAnnotation("List organizations accessible to the authenticated Terraform user"),
+			mcp.WithReadOnlyHintAnnotation(true),
+			mcp.WithOpenWorldHintAnnotation(false),
+			mcp.WithString("query", mcp.Description("Optional: Filter organizations by name or email")),
+		), func(ctx context.Context, request mcp.CallToolRequest) (*mcp.CallToolResult, error) {
+			var query string
+			if queryVal, ok := request.Params.Arguments["query"]; ok {
```
BBBmau, line 26: nice that the endpoint had a query parameter! We should consider changing the name of the tool to `searchOrganmizations` because of it.

pkg/hashicorp/tfenterprise/organization/handlers.go
```diff
@@ -0,0 +1,89 @@
...
+					Query: query,
+				}
+				orgList, err := tfeClient.Organizations.List(ctx, options)
+				if err != nil {
+					return nil, util.LogAndWrapError(logger, "listing organizations", err)
+				}
+				allOrgs = append(allOrgs, orgList.Items...)
+				totalCount = orgList.TotalCount
+
+				if len(allOrgs) >= totalCount || len(orgList.Items) == 0 {
```
BBBmau: Why is it necessary to check `len(allOrgs) >= totalCount`? Wouldn't this always trigger on the first iteration since we are comparing the totalCount of the list with the same length list of `orgList.Items`?

pkg/hashicorp/tfenterprise/organization/utils.go
```diff
@@ -0,0 +1,148 @@
...
+	return builder.String(), nil
+}
+
+// formatOrganization returns a formatted summary of a TFE Organization.
+func formatOrganization(org *tfe.Organization) string {
+	if org == nil {
+		return "Organization: <nil>"
+	}
+
+	return fmt.Sprintf(`Organization Details
```
BBBmau, line 52: this formatting is nice!

BBBmau (commented):
I've tested this locally, it works great however I can't truly test the pagination feature which is where my comments are coming from.

pkg/hashicorp/tfenterprise/organization/handlers.go
aditya2548: `len(allOrgs) >= totalCount` is used to determine when we've collected all organizations across paginated responses. Since `orgList.Items` only contains the current page's results, we need to track the total across all pages. This check ensures we stop once we've fetched everything based on the totalCount value returned by the API.

The `len(orgList.Items) == 0` check is an additional safeguard in case the API doesn't return the full count or if something unexpected happens (like a partial page).

pkg/hashicorp/tfenterprise/organization/handlers.go
aditya2548: Good point, done!

pkg/hashicorp/tfenterprise/organization/handlers.go
BBBmau: ah okay so allOrgs just represents what's on the page while totalCount is the total organization count across all pages.

Can we include a comment on this or maybe rename to be clearer?

pkg/hashicorp/tfenterprise/organization/handlers.go
```diff
@@ -0,0 +1,89 @@
...
+				if queryStr, ok := queryVal.(string); ok {
+					query = queryStr
+				}
+			}
+
+			var allOrgs []*tfe.Organization
+			pageNumber := 1
+			totalCount := 0
+
+			// Iterate through all pages to collect all organizations
```
BBBmau, line 36: this should be done in a separate PR but we should look into how we can have single pagination implementation for use across all tools since I imagine we'll be running into pagination constantly.

Edit: created a issue [here](https://github.com/hashicorp/terraform-mcp-server/issues/65#issue-3081604555)

pkg/hashicorp/tfenterprise/organization/handlers.go
aditya2548: Updated variable names to make it clearer.

PR #75, opened 2025-05-27
fix: no declared logging capability which actually not support yet

### LINKED ISSUE #74: [Bug]: The server actually not support "Logging" capability as it declared
I use client tools [mcp-cli-host](https://github.com/vincent-pli/mcp-cli-host),
it will try to set log level if server declared supporting, but get the error:
`Error initializing server terraform: Method logging/setLevel not found`

vincent-pli:
fix #74

### DIFF
--- a/cmd/terraform-mcp-server/main.go
+++ b/cmd/terraform-mcp-server/main.go
@@ -62,7 +62,6 @@ func NewServer(version string, opts ...server.ServerOption) *server.MCPServer {
 	defaultOpts := []server.ServerOption{
 		server.WithToolCapabilities(true),
 		server.WithResourceCapabilities(true, true),
-		server.WithLogging(),
 	}
 	opts = append(defaultOpts, opts...)
 


BBBmau (commented):
can you link where this is documented as not supported? It looks to be supported based on the mark3labs source code: https://github.com/mark3labs/mcp-go/blob/c7c0e13ad2675746f78b2580ad4a8de57b2bbbc6/server/server.go#L266-L270

vincent-pli:
@BBBmau 
Yes, the `mcp-go` use 
```
func WithLogging() ServerOption {
	return func(s *MCPServer) {
		s.capabilities.logging = mcp.ToBoolPtr(true)
	}
}
```
to declared itself support `logging`, it add "logging = true" in `capability` of server  to  a `mcp client` (for example [mcp_cli_host](https://github.com/vincent-pli/mcp-cli-host))

but actually, to really support `logging`, the server must support two function:
- logging/setLevel (this function is for client, since client will call the function to set the level of log)
- notifications/message (this function is called by server itself, to send notification to client, as I mentioned in #76)

but we actually not have such thing, so should not declared supported(by call `WithLogging`) 

please check the specification: https://modelcontextprotocol.io/specification/2025-03-26/server/utilities/logging

BBBmau:
ah so from my understanding is although we support logging on server side we would need client support which is something we don't have control over. Thanks for the explanation, i'll be merging this in and looking into the second issue you made.

PR #80, opened 2025-06-04
update base image to scratch

mukeshjc:
the CVEs mentioned are currently present in the latest version of alpine images and busybox releases.

updating the image to `scratch` to reduce the vulnerability surface area.

https://security.alpinelinux.org/branch/edge-main (look for busybox related CVEs)

---
main CVE's reported during CI execution

image

PR #81, opened 2025-06-05
Changing base container image to Scratch

gautambaghel:


Dockerfile
```diff
@@ -14,27 +11,21 @@
...
-# Make a stage to run the app
-FROM docker.mirror.hashicorp.services/alpine:3.21 AS dev
+FROM scratch AS dev
 ARG VERSION="dev"
-# Set the working directory
 WORKDIR /server
-# Copy the binary from the build stage
 COPY --from=devbuild /build/terraform-mcp-server .
-# Command to run the server
+COPY --from=devbuild /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
```
mukeshjc: we need this same statement added to the `release-default` build target below. so it would be added after the copy of binary is done at line 48.

Dockerfile
```diff
@@ -43,9 +34,7 @@ CMD ["./terraform-mcp-server", "stdio"]
 #
 # ===================================
 
-# default release image (refereced in .github/workflows/build.yml)
-# -----------------------------------
```
mukeshjc: i would suggest to keep this comment in the document. it's helpful to know that build.yml file uses the below docker image build definition which actually lands up in releases.

Dockerfile
```diff
@@ -1,9 +1,6 @@
 # Copyright (c) HashiCorp, Inc.
 # SPDX-License-Identifier: MPL-2.0
 
-# This Dockerfile contains multiple targets.
-# Use 'docker build --target=<name> .' to build one.
```
mukeshjc: this comment also seems helpful to keep, for folks who may not know how multi-stage builds and multi-target Dockerfiles work.

Dockerfile
gautambaghel: done!

Dockerfile
gautambaghel: done!

gautambaghel:
closing in favor of #80

PR #82, opened 2025-06-05
`unit-test`: add `TestLogAndReturnError`

BBBmau:
Figured there were more unit tests to add but we only were missing `TestLogAndReturnError`


gautambaghel:
@BBBmau - can you update the out-of-date base branch? otherwise looks good to me!

Issue #90, opened 2025-06-13
SearchModules doesn't support provider, namespace, or verified parameters


KiPIDesTAN:
The Terraform MCP server's SearchModules tool supports both the Terraform Registry API for List Modules (when moduleQuery is empty) and Search Modules (when moduleQuery is populated) API call. However, the tool is missing support for the List Module API's provider and verified parameters. The tool is also missing support for the Search Modules API's provider, namespace, and verified parameters. The API definitions are listed on the [Terraform Registry API](https://developer.hashicorp.com/terraform/registry/api-docs) page.

I've already made a fix and will be submitting a PR momentarily. I just wanted to make the issue so it's linked appropriately.

PR #91, opened 2025-06-13
SearchModules tool support for provider, namespace, and verified parameters

KiPIDesTAN:
This pull request enhances the functionality of the `searchModules` feature by adding support for new query parameters (`provider`, `namespace`, and `verified`) and updating the corresponding test cases to validate these changes. The most important changes include modifications to the `searchModules` function, updates to the `SearchModules` handler, and the addition of new test cases.

This change is in alignment with the List Modules and Search Modules [Terraform Registry API](https://developer.hashicorp.com/terraform/registry/api-docs) and fixes issue #90 .

### Enhancements to `searchModules` functionality:

* [`pkg/hashicorp/tfregistry/utils.go`](diffhunk://#diff-8de9e331f18fda531b9a3c0e6bfaf5007af711073524f9bc542b63b822fa9378L260-R275): Updated the `searchModules` function to accept and process new parameters (`provider`, `namespace`, and `verified`). These parameters are appended to the query string when constructing the API request.

* [`pkg/hashicorp/tfregistry/handlers.go`](diffhunk://#diff-ae9e134989cfa5b183e0c671891479217f42eadae57e07077a678f226785ead5R152-R193): Updated the `SearchModules` handler to include the new parameters (`provider`, `namespace`, and `verified`) in the request arguments. Default values and type checks for these parameters were also added.

### Test case additions:

* [`e2e/payloads.go`](diffhunk://#diff-c8e7cb868f756fa2ce58f0d66b03fe5ca3d4ee3c520e8c6bf5dca5749184b251R272-R323): Added multiple test cases to validate the `searchModules` functionality with various combinations of the new parameters (`provider`, `namespace`, and `verified`). These include scenarios with empty queries, specific providers, namespaces, and verified flags.


pkg/hashicorp/tfregistry/handlers.go
```diff
@@ -149,19 +149,48 @@ func SearchModules(registryClient *http.Client, logger *log.Logger) (tool mcp.To
...
+			mcp.WithString("provider",
+				mcp.Description("Terraform provider to limit the search on."),
+				mcp.DefaultString(""),
+			),
+			mcp.WithString("namespace",
+				mcp.Description("Terraform namespace to limit the search on."),
+				mcp.DefaultString(""),
+			),
+			mcp.WithBoolean("verified",
+				mcp.Description("True when searching only for parnter modules. False, otherwise."),
```
thrashr888:
```suggestion
				mcp.Description("True when searching only for partner modules. False, otherwise."),
```

KiPIDesTAN:
@gautambaghel , you're free to not accept the PR, but I would encourage you to revisit your assumptions and design. I've had far more success getting the right response with fewer agent calls by including provider, namespace, and verified as _optional_ parameters. "AI sometimes fumbles with calls with additional parameters" is the opposite of my experience. I've been using this solution for a couple months now.

My comments below are only so you have a record for the future. 

As a basic example, "Write terraform that creates an Azure resource group using only Azure Verified Modules." This prompt should allow an agent to quickly identify Azure/avm-res-resources-resourcegroup/azurerm/0.2.1.

With the current MCP implementation, the agent will call SearchModules and use variations of "resource group" as its query. The registry v1 search API only supports basic keyword search on limited fields within the module meta-data. This is part of the issue.

If we search "resource group", as is often the first phrase agents search, v1 search returns 15 responses, but there are well over 105 that might require paginating through. In this situation, the most common paths taken by models are the following:

1. The model picks the "best" module from the 15 responses, even if the best module isn't even close to what we are interested in. Read "This module must be right because it has the most downloads, even though none of the other criteria fits."
2. The model doesn't like any of the responses and tries a new term to search.

I think there's a third expected path by the existing design, which is that the model will paginate through the search results because there is a currentOffset parameter in the SearchModule function. However, this pagination _never_ occurs. The model does option 2 because the MCP server's response to the agent never makes an indication that there are more possible results and never mentions an offset. This means currentOffset is never used.

This highlights two issues with the current implementation.
(1) Search is basic and gives poor responses.
(2) Even if the search term potentially contains the module requested, the MCP server doesn't provide enough information back to the agent to review the fully returned results.

As for "AI sometimes fumbles with calls with additional parameters", this hasn't been my experience. Models driving agents see "optional" on parameters and more often than not err on the side of caution, not populating values if they're not highly confident of the desire.

For example, if we include the namespace and provider as parameters (my PR), models will typically not populate these values if my example is "Write Terraform that creates an Azure resource group using only Azure Verified Modules." However, it will populate namespace and provider if I say "Write Terraform that creates an Azure resource group using only Azure Verified Modules. Azure Verified Modules only come from the Azure namespace and azurerm provider." 

To that end, you could make the same argument about namespace and provider that you made about verified modules - "you can...force [the] decision by asking it to only pick verified modules." However, if you're going to be explicit about wanting things that can be filtered out during search to give the agent a more accurate, smaller list of modules to choose, why wouldn't you?

The approach of "just put what you want explicitly in the prompt and have the model filter it out later" puts you in the situation of requiring the model to process more data, consume tokens, cost more money, and potentially still miss a lot of return results because it doesn't know it needs to paginate through more search results.

PR #93, opened 2025-06-18
add ECR release support and update 'latest' tag of dev images

mukeshjc:
ECR release support. 

Backporting this change to `release/0.1.0` is tricky, because that would be a new hash and we would have to re-upload all artifacts for 0.1.0 including to Dockerhub as well. Choosing not to do that for 0.1.0 and we start with 0.2.0 on ECR. It's not clean but folks should ideally be using just the `:latest` tag mostly when pulling the image.


gautambaghel (approved):
LGTM!

PR #122, opened 2025-07-31
Feature/version tools - adding 2 new tools, get latest provider and module versions

gautambaghel:
## PCI review checklist

### DIFF
--- a/pkg/tools/get_latest_provider_version.go
+++ b/pkg/tools/get_latest_provider_version.go
@@ -1,0 +1,54 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package tools
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/hashicorp/terraform-mcp-server/pkg/client"
+	"github.com/hashicorp/terraform-mcp-server/pkg/utils"
+	"github.com/mark3labs/mcp-go/mcp"
+	"github.com/mark3labs/mcp-go/server"
+	log "github.com/sirupsen/logrus"
+)
+
+// GetLatestProviderVersion creates a tool to get the latest provider version from the public registry.
+func GetLatestProviderVersion(registryClient *http.Client, logger *log.Logger) server.ServerTool {
+	return server.ServerTool{
+		Tool: mcp.NewTool("get_latest_provider_version",
+			mcp.WithDescription("Fetches the latest version of a Terraform provider from the publi registry"),
+			mcp.WithTitleAnnotation("Get Latest Provider Version"),
+			mcp.WithOpenWorldHintAnnotation(true),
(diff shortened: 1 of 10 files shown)


pkg/tools/get_latest_provider_version.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/mark3labs/mcp-go/mcp"
+	"github.com/mark3labs/mcp-go/server"
+	log "github.com/sirupsen/logrus"
+)
+
+// GetLatestProviderVersion creates a tool to get the latest provider version from the public registry.
+func GetLatestProviderVersion(registryClient *http.Client, logger *log.Logger) server.ServerTool {
+	return server.ServerTool{
+		Tool: mcp.NewTool("get_latest_provider_version",
+			mcp.WithDescription("Fetches the latest version of a Terraform provider from the publi registry"),
```
thrashr888:
```suggestion
			mcp.WithDescription("Fetches the latest version of a Terraform provider from the public registry"),
```

pkg/tools/get_latest_provider_version.go
```diff
@@ -0,0 +1,54 @@
...
+				mcp.Description("The name of the provider, e.g., 'aws'")),
+		),
+		Handler: func(ctx context.Context, req mcp.CallToolRequest) (*mcp.CallToolResult, error) {
+			return getLatestProviderVersionHandler(registryClient, req, logger)
+		},
+	}
+}
+
+func getLatestProviderVersionHandler(registryClient *http.Client, request mcp.CallToolRequest, logger *log.Logger) (*mcp.CallToolResult, error) {
+	namespace, err := request.RequireString("namespace")
```
thrashr888, line 40: when i built mine, i was wondering if we should just default to `hashicorp` since that would be most often correct if the llm doesn't know it.

i also wonder if we want to lower-case all the input to normalize it. should test for this too.

thrashr888 (dismissed):
lgtm, with some nits

pkg/tools/get_latest_provider_version.go
gautambaghel: defaulting to hashicorp stops the model from thinking and ends up calling the tools twice, it's better to add it in the description (saying e.g, hashicorp - which we've done) that way if it has no other choice it uses hashicorp. hashicorp owns <20 providers, we got at least 200 decent providers out there

the APIs are case insensitive, but should be easy to add

pkg/tools/get_latest_provider_version.go
gautambaghel: fixed

pkg/tools/get_latest_provider_version.go
thrashr888: good points. totally agree the description covers it well enough.

PR #152, opened 2025-08-26
Feature/workspaces

gautambaghel:
## PCI review checklist


pkg/tools/tfe/static/default-cli-run.md
```diff
@@ -0,0 +1,26 @@
...
+    workspaces { 
+      name = "gcp-oidc-0x7a" 
+    } 
+  } 
+}
+```
+
+
+3. Run terraform init to initialize the workspace.
+1. Run terraform apply to start the first run for this workspace.
```
mayank-sisodiya:
```suggestion
4. Run terraform apply to start the first run for this workspace.
```

pkg/tools/tfe/static/default-cli-run.md
```diff
@@ -0,0 +1,26 @@
+### CLI-driven runs
```
mayank-sisodiya, line 1: Are these steps mandatory?

pkg/tools/tfe/delete_workspace_safely.go
```diff
@@ -0,0 +1,71 @@
+// Copyright (c) HashiCorp, Inc.
```
mayank-sisodiya, line 1: file name can be just `delete_workspace`.

pkg/tools/tfe/search_workspaces.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/hashicorp/terraform-mcp-server/pkg/client"
+	"github.com/hashicorp/terraform-mcp-server/pkg/utils"
+	log "github.com/sirupsen/logrus"
+
+	"github.com/mark3labs/mcp-go/mcp"
+	"github.com/mark3labs/mcp-go/server"
+)
+
+// SearchWorkspaces creates a tool to search for Terraform workspaces.
+func SearchWorkspaces(logger *log.Logger) server.ServerTool {
```
mayank-sisodiya: We should name the tool as `ListWorkspaces` following the [API endpoints](https://developer.hashicorp.com/terraform/enterprise/api-docs/workspaces#list-workspaces).

gautambaghel: I do appreciate the idea of keeping things consistent but this is where most companies get MCP wrong, MCP != APIs or SDKs, the idea is not to have a comprehensive way to work with the product, it's really about what are the most best way AI can combine the tools to perform a particular task, it could be multiple different APIs combined or it could just be a utility function that might be very useful
 
I think `listing` as a function for MCP is very useless, I can't imagine scenarios where the MCP listing everything is useful, it wastes a lot of tokens as well. `searching` is much more efficient, it forces the model to find out exactly what users were looking for and pass that as the query argument. The response is digestible too, doesn't consume too many tokens, don't have to paginate to find what to look etc. etc.

Tests.md
```diff
@@ -0,0 +1,99 @@
+# Running tests
```
mayank-sisodiya: Are these the instructions to run e2e tests?

pkg/tools/tfe/delete_workspace_safely.go
gautambaghel: We aren't adding force delete workspace  as a tool yet (I don't thing we even should) but there's two kinds of deletes in HCP Terraform,

* [Safe delete](https://developer.hashicorp.com/terraform/cloud-docs/api-docs/workspaces#safe-delete-a-workspace) - deletes the workspace only if there are no resources
* [Force delete](https://developer.hashicorp.com/terraform/cloud-docs/api-docs/workspaces#force-delete-a-workspace) - delete the workspace even if active resources

If the model choses to delete a workspace then the `safely_delete` qualifier is good to have

Tests.md
```diff
@@ -0,0 +1,99 @@
...
+      ```
+
+### Required ENVVARS
+
+Tests are run against an actual backend so they require a valid backend address and token:
+
+1. `TFE_ADDRESS` - URL of a HCP Terraform or Terraform Enterprise instance to be used for testing, including scheme. Example: `https://tfe.local`
+1. `TFE_TOKEN` - A [user API token](https://developer.hashicorp.com/terraform/cloud-docs/users-teams-organizations/users#tokens) for the HCP Terraform or Terraform Enterprise instance being used for testing.
+
+**Note:** Alternatively, you can set `TFE_HOSTNAME` which serves as a fallback for `TFE_ADDRESS`. It will only be used if `TFE_ADDRESS` is not set and will resolve the host to an `https` scheme. Example: `tfe.local` => resolves to `https://tfe.local`
```
mayank-sisodiya: I don't see any occurrence of TFE_HOSTNAME is this repo, where is this fallback handled?

Tests.md
gautambaghel: No, this is to test with actual HCP Terraform tokens, I haven't added live testing yet (only in/out mock tests), but we should start adding live HCP TF tests to make sure tools are running they way they're supposed to

it should be similar to `tfe-go`

Tests.md
gautambaghel: `tfe-go` automatically picks it up I believe

pkg/tools/tfe/search_workspaces.go
mayank-sisodiya: 1. This is a very standard way of doing these where you have a list API with filters to restrict the result based upon the need and should not be other way round. Is naming the tool as `search` than `list` going to give better results?? I just think the LLM would be able to comprehend better if the tools also follows on the best practises followed for years(training data).
2. `listing as a function for MCP is very useless` - can you elaborate more on this. Even with current implementation we are not avoiding this if pass no search criteria as stated in the description.

pkg/tools/tfe/delete_workspace_safely.go
mayank-sisodiya: For now this implementation holds good.

pkg/tools/tfe/static/default-cli-run.md
gautambaghel: These are simply taken from HCP Terraform UI when you create an empty workspace on how to get started

Tests.md
gautambaghel: local testing for now but we should add them to GitHub Actions

pkg/tools/tfe/search_workspaces.go
gautambaghel: A typical customer would have ~1000 workspace, in what context would listing all of those be useful? especially with pagination when the AI client keep going trying to find the appropriate workspace.

naming the tool this way and adding this description forces the LLM to think better

In the current implementation we're not avoiding it I just think that naming the tool `search` (and adding what I added to the description) forces the model to be more specific with it's request rather than simply "listing" workspaces. The more specific we are the better the results, fewer the tokens needed.

As far as trainings, LLMs are also very aware of search functions I'm not sure what gives list functions the edge.

gautambaghel: look at github for example, they have list [repositories APIs](https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#list-repositories-for-the-authenticated-user) obv but they've implemented [search](https://github.com/github/github-mcp-server/blob/main/pkg/github/search.go) instead


~~they don't have  any [list tools](https://github.com/github/github-mcp-server/tree/main/pkg/github) for that matter~~, in our case we have list org and list projects because their quantity would ~ <100

gautambaghel: I take that back, they DO have list implemented, they have both search and list (which also leads to tool bloating tbh)

Alright if the tool name being `list` seems important to you I'll change it

gautambaghel: Do you think I should name it `list_terraform_workspaces` to be more obvious?

I would need to add it to all the variations of the tool, `update_workspace` -> `update_terraform_workspace` etc. etc.

pkg/tools/tfe/static/default-cli-run.md
mayank-sisodiya: Steps indicate running Terraform CLI commands manually, is terraform CLI installation a prerequisite for TFE tools? We are expecting TFE token and TFE Address, so terraform login shouldn't be required, is my understanding right?

mayank-sisodiya (approved):

PR #155, opened 2025-08-29
Feature/rate-limiting

gautambaghel:
## PCI review checklist


dduzgun-security (dismissed):
LGTM, nice improvement 👍

mayank-sisodiya (approved):

gautambaghel:
> Do we have the flexibility to disable rate limiting entirely? Considering that the server will be run locally, should we give the option. This can be followup and not a blocker for this PR.

Yeah, this was discussed a bit here: https://github.com/hashicorp/vault-mcp-server/pull/49#issuecomment-3282086999

We can do as you suggest regardless but having a big default buffer of 50 API calls per second for burst should be good enough for now

PR #156, opened 2025-08-29
Adding instructions for MCP servers and examples for AI clients

gautambaghel:
## PCI review checklist


cmd/terraform-mcp-server/instructions.md
```diff
@@ -0,0 +1,76 @@
+# Terraform MCP Server Tool Hints
+
+The Terraform MCP server provides tools for generating better Terraform code through registry integration and automating workflows via HCP Terraform/Enterprise APIs.
+
+## Tool Usage Guidelines
+
+**BEFORE generating any Terraform code**: Query registries for latest provider/module versions and styling guidelines. When TFE_TOKEN exists, search private registries first, then fall back to public.
+**Provider Consistency**: All modules in a project must use compatible provider versions. Verify with get_provider_details before generating code.
+**Validation Flow**: Run terraform validate immediately after generation, then terraform plan only if validation passes. Use terraform fmt to format code as needed.
+**User Confirmation Required**: ALWAYS get explicit yes/no confirmation before: create_run, apply_run, discard_run, cancel_run.
```
kartheek-hc: nit: It looks like these aren't being rendered in the markdown as separate lines but I don't think it matters to the Agent/model

cmd/terraform-mcp-server/instructions.md
```diff
@@ -0,0 +1,76 @@
+# Terraform MCP Server Tool Hints
+
+The Terraform MCP server provides tools for generating better Terraform code through registry integration and automating workflows via HCP Terraform/Enterprise APIs.
+
+## Tool Usage Guidelines
+
+**BEFORE generating any Terraform code**: Query registries for latest provider/module versions and styling guidelines. When TFE_TOKEN exists, search private registries first, then fall back to public.
```
kartheek-hc: Just curious, does the Agent consuming these instructions know whether the MCP server has been setup with a Terraform token? I wonder if we should just say something like
```
When private registry tools are available, prefer them first, then fallback to public
```

instructions/example-copilot-instructions.md
```diff
@@ -0,0 +1,134 @@
...
+- `main.tf` – Primary resource and data source definitions. Must exist in every module, even if empty.
+- `variables.tf` – Input variable definitions (alphabetical order). Must exist in every module, even if empty.
+- `outputs.tf` – Output value definitions (alphabetical order). Must exist in every module, even if empty.
+- `README.md` – Required in the root module. Describes the module's purpose, usage, and may include diagrams or examples.
+
+### Recommended Module Structure
+- `LICENSE` – Recommended in the root of the module, especially for public modules, to clarify usage rights.
+- `providers.tf` – Provider configurations and requirements.
+- `terraform.tf` – Terraform version and provider requirements.
+- `backend.tf` – Backend configuration for state storage.
```
kartheek-hc: nit: Should we say `backend.tf` is only needed for root modules?

instructions/example-copilot-instructions.md
```diff
@@ -0,0 +1,134 @@
...
+## Resource Organization
+- Define data sources before the resources that reference them to build logically
+- Group related resources together (networking, compute, storage)
+- Order resource parameters consistently: meta-arguments, resource-specific parameters, nested blocks, lifecycle, depends_on
+
+## Resource Naming
+- Use descriptive nouns separated by underscores
+- Do not include the resource type in the resource name
+- Wrap resource type and name in double quotes
+- Example: `resource "aws_instance" "web_server" {}` not `resource aws_instance webserver_instance {}`
```
kartheek-hc: nit:
```suggestion
- Example: `resource "aws_instance" "web_server" {}` not `resource "aws_instance" "webserver_instance" {}`
```

instructions/example-copilot-instructions.md
```diff
@@ -0,0 +1,134 @@
+---
+applyTo: "**/*.{tf,hcl}"
+description: "HashiCorp style guidelines for writing Terraform code"
+---
+
+# Terraform Code Style Guidelines
```
kartheek-hc, line 6: These instructions are really good

kartheek-hc (dismissed):

cmd/terraform-mcp-server/instructions.md
devops-adeel: **second nit:**  (rather, a genuinely curious question) Since agents are very receptive to Markdown, will the presence of backticks wrapped around tool names provide a stronger indication that they are indeed tools or commands? 

e.g. `create_run` instead of create_run.

cmd/terraform-mcp-server/instructions.md
```diff
@@ -0,0 +1,76 @@
...
+2. create_run → get_run_details to monitor
+3. get_plan_details/logs to review changes
+4. User confirmation → apply_run OR discard_run
+
+**Variable Configuration**:
+1. search_workspace_variables to check existing
+2. create/update_workspace_variable as needed
+3. For multi-workspace: create_variable_set → attach_variable_set_to_workspaces
+
+## Error Handling
```
devops-adeel, line 75: 🔥 🔥 🔥

instructions/example-copilot-instructions.md
```diff
@@ -0,0 +1,134 @@
...
+
+This project follows HashiCorp's official Terraform style guide for consistent, maintainable infrastructure-as-code.
+
+## Project Context
+Terraform configurations define infrastructure resources using HashiCorp Configuration Language (HCL). 
+These instructions ensure code consistency, readability, and maintainability across all Terraform files.
+
+## Module and Repository Structure
+
+Organize your Terraform modules and repositories as follows:
```
devops-adeel, line 17: Don't take this as a suggestion: Just me thinking out loud 💭 .

I wonder if the LLM/Agent would be even more receptive to an example of a directory tree. 🤔 

example from our [public page](https://developer.hashicorp.com/terraform/language/modules/develop/structure).

```
.
├── README.md
├── main.tf
├── variables.tf
├── outputs.tf
├── ...
├── modules/
│   ├── nestedA/
│   │   ├── README.md
│   │   ├── variables.tf
│   │   ├── main.tf
│   │   ├── outputs.tf
│   ├── nestedB/
│   ├── .../
├── examples/
│   ├── exampleA/
│   │   ├── main.tf
│   ├── exampleB/
│   ├── .../
```

instructions/example-copilot-instructions.md
```diff
@@ -0,0 +1,134 @@
...
+- `backend.tf` – Backend configuration for state storage.
+- `locals.tf` – Local value definitions.
+- `modules/` – Local module definitions in `./modules/<module_name>`. Nested modules with a `README.md` are considered public; those without are considered internal.
+- `examples/` – Directory containing usage examples for the module and submodules. Each example may have its own README. Example `module` blocks should use the external source address, not a relative path.
+
+### Additional Structure Guidance
+- Split large configurations into logical files (e.g., `network.tf`, `compute.tf`, `storage.tf`).
+- Keep module repositories focused on single infrastructure concerns.
+- Use three-part naming for module repositories: `terraform-<PROVIDER>-<NAME>`.
+- Store local modules in `./modules/<module_name>`.
```
MrFixit96, line 59: Should we have an explanation for when to use local modules?

instructions/example-copilot-instructions.md
```diff
@@ -0,0 +1,134 @@
...
+- Group related resources together (networking, compute, storage)
+- Order resource parameters consistently: meta-arguments, resource-specific parameters, nested blocks, lifecycle, depends_on
+
+## Resource Naming
+- Use descriptive nouns separated by underscores
+- Do not include the resource type in the resource name
+- Wrap resource type and name in double quotes
+- Example: `resource "aws_instance" "web_server" {}` not `resource aws_instance webserver_instance {}`
+
+## Variables and Outputs
```
MrFixit96, line 87: Should we have an instruction to specifically look to forward the output of a module to the output of the consumer? (ie. we always want to share the ip address of an ec2 instance with the user and not make them hunt for it)

cmd/terraform-mcp-server/instructions.md
gautambaghel: yeah, backticks might be helpful, I'll add them. newlines might inmprove the readability for humans which is also necessary ;) i'll add them

cmd/terraform-mcp-server/instructions.md
gautambaghel: you're right they might not be explicitly aware of the token

instructions/example-copilot-instructions.md
gautambaghel: maybe even a good insertion point for TFC ;) and say if unavailable fall back to other state storages

instructions/example-copilot-instructions.md
```diff
@@ -0,0 +1,134 @@
...
+- Set `sensitive = true` for passwords and private keys
+- Order variable parameters: type, description, default, sensitive, validation blocks
+- Order output parameters: description, value, sensitive
+- Use descriptive names with underscores to separate words
+
+## Comments and Documentation
+- Use `#` for single and multi-line comments (not `//` or `/* */`)
+- Write self-documenting code; use comments only to clarify complexity
+- Add comments above resource blocks to explain non-obvious business logic
+
```
MrFixit96, line 100: I don't know which section this would belong to, but when a TFE_TOKEN is present, we should look to include the use of tagging as much as possible for resources as a best practice.

instructions/example-copilot-instructions.md
```diff
@@ -0,0 +1,134 @@
...
+- Use local values sparingly to avoid making code harder to understand
+- Define in `locals.tf` if referenced across multiple files
+- Define at the top of a file if specific to that file only
+- Use descriptive nouns with underscores for local value names
+
+## Dynamic Resource Management
+- Use `count` for nearly identical resources
+- Use `for_each` when resources need distinct values that cannot be derived from integers
+- Use `count` with conditional expressions for optional resources: `count = var.enable_feature ? 1 : 0`
+- Use meta-arguments sparingly and add comments when the effect is not obvious
```
MrFixit96: let's actually add comments to explain the effect regardless of if it's obvious. one of the ethos' of Terraform HCL is that it should be explicitly explicit and never leaving you to decide what it might be doing.

instructions/example-copilot-instructions.md
gautambaghel: maybe a new line in itself, what would the default tags be? or should we not specify

cmd/terraform-mcp-server/instructions.md
gitrgoliveira: Maybe it needs a bit more direct instruction, so it always tries the private registry first?
Something like:
BEFORE generating any Terraform code: Query the private registry, with <name_of_tool> for latest provider/module versions and styling guidelines. If the call fails or has no result use the public registry with <name_of_tool>.

instructions/example-copilot-instructions.md
```diff
@@ -0,0 +1,134 @@
...
+
+## Provider Configuration
+- Always include a default provider configuration
+- Define all providers in the same file (`providers.tf`)
+- Define the default provider first, then aliased providers
+- Use `alias` as the first parameter in non-default provider blocks
```
MrFixit96, line 117: Should we have comment on when to use an alias?

cmd/terraform-mcp-server/instructions.md
```diff
@@ -0,0 +1,76 @@
...
+- Registry failures: Try private first (if token), fallback to public
+- Run failures: Check get_run_details, get_plan_details and logs before retry
+- Variable conflicts: search_workspace_variables first to avoid duplicates
+
+## Security Notes
+- Never expose TFE_TOKEN in outputs
```
gitrgoliveira: The agent will never have direct access to it though, right?

cmd/terraform-mcp-server/instructions.md
```diff
@@ -0,0 +1,76 @@
...
+- Use delete_workspace_safely to prevent accidental resource deletion
```
gitrgoliveira: maybe the agent shouldn't even have access to do this?
reading the description of `delete_workspace_safely` this shouldn't be a possible scenario?

instructions/example-copilot-instructions.md
devops-adeel, line 1: @gautambaghel - I tell you what I find interesting with this document. I'm assuming this is an example system prompt that a user would put in co-pilot, right? If so....how effective is this (i.e. hoping for the end user to copy this in to setup) vs adding instructions in mcp `init`, instructing it to use the [mcp resource](https://github.com/hashicorp/terraform-mcp-server/blob/df04fe0a97a57d43ea00db04dd8e03c6fe715261/pkg/resources/resources.go#L30) for `/terraform/style-guide`?

I would think we'd have better control of it's behaviour if we asked it to use the resource?

Keen to hear your thoughts @MrFixit96

gautambaghel: I fundamentally agree with you, however a lot of AI clients haven't implemented resource properly, unless you feel otherwise? just from my limited experimentation

exposing this file to me is really about showing customers patterns on what they could do, because every customer has their own patterns and practices.

cmd/terraform-mcp-server/instructions.md
emrousselle: > When private registry tools are available, prefer them first, then fallback to public

Overall I'm fine with this statement, but in some cases the user is using the private registry to whitelist content from the public registry. Sentinel policies then are used to ensure that the Terraform code only uses the private registry.

In that scenario, if the MCP server recommends public registry providers/modules because no suitable alternative was found in the private registry, the practitioner will end up with code that'll get rejected at the policy-check gate.

I'm not sure there is an easy way to handle this situation, unless the MCP server can consume "organization-specific" instructions that specify the private registry should be the **only** source of providers/modules.

gautambaghel: @emrousselle - yeah that's why there are also example instruction files attached in this PR, if there are some custom rules users want to implement around their own workflows they can do that

instructions/example-mcp-instructions.md
```diff
@@ -0,0 +1,113 @@
...
+## Best Practices for LLM Implementation
+
+### 1. Context Preservation
+- Maintain state of previous MCP server queries within the session
+- Track which registries have been searched
+- Remember user preferences and requirements stated earlier
+
+### 2. Progressive Enhancement
+- Start with minimal viable configuration
+- Iteratively add complexity based on validation results
```
MrFixit96, line 108: Should we or can we introduce Test Driven Development here and actually create the tests we want to achieve before writing code, then improving, validating, iterating?

This should increase accuracy in the end and hopefully cut down the overall number of iterations.

cmd/terraform-mcp-server/instructions.md
gautambaghel: that's fair, let me edit this

instructions/example-copilot-instructions.md
gitrgoliveira: Perhaps we can have some example tags, but the user will really need to edit this section when they place it into their repo.
I think that whatever tags we decide to use, they will always be wrong for someone, but a placeholder at least allows the user to edit the code?

instructions/example-mcp-instructions.md
gitrgoliveira: honestly.. I tried that and `terraform test` and it kept breaking the LLM and the code, because `terraform test` doesn't implement mocks and other features properly. Example: doesn't support ephemeral resources.

instructions/example-copilot-instructions.md
emrousselle: An approach that I've used is to add tags identifying the Project and Workspace that manages the resource, and a tag indicating that its managed by Terraform.

That way, if I'm in the console looking at the resource, I know where to look for if I need to change it.

instructions/example-copilot-instructions.md
devops-adeel: or even in this document too, have you tested to see copilot behaviour if you instructed it to use resources `PostTool`?

devops-adeel: > I fundamentally agree with you, however a lot of AI clients haven't implemented resource properly, unless you feel otherwise? just from my limited experimentation
> 
> exposing this file to me is really about showing customers patterns on what they could do, because every customer has their own patterns and practices.

Hey @gautambaghel - that is a fair comment, I agree with you on not over-engineering this on the server-side, it isn't worth it.

However, just going by [this feature matrix](https://modelcontextprotocol.io/clients#feature-support-matrix) published by Anthropic (showing `VS Code GitHub Copilot` supporting `resources`, I would be curious how :copilot:  will behave if we added instructions here about using `resources`.

@MrFixit96 - would you do the honours? 😄

instructions/example-mcp-instructions.md
```diff
@@ -0,0 +1,113 @@
...
+**ALWAYS** consult the MCP server before generating any Terraform code to:
+1. Retrieve latest provider documentation and constraints
+2. Access organization-specific styling guidelines
+3. Identify available modules and their requirements
+4. Understand enterprise-specific policies and compliance requirements
+
+### Registry Search Priority
+When enterprise tools are enabled AND a Terraform token is provided:
+1. **First**: Search private registries for modules and providers
+2. **Second**: If no results found in private registries, fall back to public registry
```
devops-adeel, line 42: @gitrgoliveira @MrFixit96 @kartheek-hc @gautambaghel - and anyone else that commented about `TFE_TOKEN` awareness. 

I think this line right here, is much simpler and far more elegant than the ambiguity of whether a `TFE_TOKEN` has been supplied. You should instruct it on `cmd/terraform-mcp-server/instructions.md` (*it's a shame you can't `@` files in the repo*) to default to `search_private_modules` regardless.

cmd/terraform-mcp-server/instructions.md
emrousselle: Not unless the practitioner is iterating locally. When doing local development, you might have a TFE_TOKEN variable exported which the agent could have access to.

It may not be common, but it's definitely a possibility.

instructions/example-copilot-instructions.md
emrousselle: > I'm assuming this is an example system prompt that a user would put in co-pilot, right?

These would be "coding instructions" added to the editor's configuration. For vs code, it lives under `.github/instructions`, for Cursor it's under` .cursorrules`. At least this is my understanding.

An argument in favor of the current approach is that the user would tailor its development environment in a way that the editor is designed to understand.

The counter argument is that it might mean that even non-IaC projects would get those instructions (depending on the user's setup), and for those projects, the TF coding instructions are irrelevant and consume space in the LLM context window.

srlynch1:
There is currently an effort to standardise on agents.md most have added support in the last week including copilot ideally we should align, rather than having copilot specific instructions

instructions/example-copilot-instructions.md
gautambaghel: I think we should add this

instructions/example-copilot-instructions.md
gautambaghel: @MrFixit96 - when DO we use it? any docs?

instructions/example-copilot-instructions.md
gautambaghel: @MrFixit96 - could you craft maybe a 1-liner for this instruction pls?

instructions/example-copilot-instructions.md
gautambaghel: agree with all of you, again these are example copilot instructions if there's something specific customers need then they can edit this and have their own version of it

instructions/example-copilot-instructions.md
gautambaghel: makes sense

instructions/example-copilot-instructions.md
gautambaghel: @MrFixit96 - sure, can you suggest edits here pls?

instructions/example-copilot-instructions.md
gautambaghel: @devops-adeel - yeah that's a great idea actually because we already have official styling guides, there's no good way to check however, which instructions works better (except of course trying both patterns manually several times)

i'm happy to replace the styling guide section for the mcp instructions though, since we covered the other scenario in the copilot instruction

instructions/example-mcp-instructions.md
gautambaghel: yup! better, i'll replace the other places where we solely mention `TFE_TOKEN`

instructions/example-copilot-instructions.md
MrFixit96: In this vein, I wanted to also ask if we plan to create a GHA to test/validate our instructions and default example do indeed make the agent behave the way we expect it to given a static set of examples to process?

gautambaghel: The issue with automated pipeline testing for prompts is that we could change nothing and tests might still fail, which could lead to frustrations, we could find a way to not make those mandatory and give a score instead of pass/fail

it will require some work though

devops-adeel: > The issue with automated pipeline testing for prompts is that we could change nothing and tests might still fail, which could lead to frustrations, we could find a way to not make those mandatory and give a score instead of pass/fail
> 
> it will require some work though

scoring is the best way - which requires it's own infra itself. I have a local setup of this: https://langfuse.com/docs/observability/overview

instructions/example-mcp-instructions.md
devops-adeel, line 1: Did we miss adding links and awareness of these docs in the `README`?

instructions/example-mcp-instructions.md
```diff
@@ -0,0 +1,113 @@
+>Note: In order to implement these custom instructions into the MCP server copy them into `cmd/terraform-mcp-server/instructions.md` and rebuild the MCP binary or docker image to use it.
```
devops-adeel, line 1: 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 

How comes I never thought about this?!?!!  Oh, I am soooo using this feature!

cmd/terraform-mcp-server/instructions.md
gautambaghel: this is resolved in the latest commit

cmd/terraform-mcp-server/instructions.md
gautambaghel: this is resolved in the latest commit

instructions/example-copilot-instructions.md
gautambaghel: this is resolved in the latest commit

instructions/example-copilot-instructions.md
gautambaghel: this is resolved in the latest commit

instructions/example-copilot-instructions.md
gautambaghel: this is resolved in the latest commit

cmd/terraform-mcp-server/instructions.md
gautambaghel: this is resolved in the latest commit

cmd/terraform-mcp-server/instructions.md
gautambaghel: this is removed in the latest commit

instructions/example-mcp-instructions.md
gautambaghel: i think i agree with ricardo here, python based tests won't work great either. imho terraform is not like a regular programming language anyway

instructions/example-mcp-instructions.md
gautambaghel: this is resolved in the latest commit

instructions/example-mcp-instructions.md
gautambaghel: nice! do it!

instructions/example-mcp-instructions.md
gautambaghel: added README, thanks for the suggestion

mayank-sisodiya (approved):

PR #179, opened 2025-09-23
Fix provider search prioritization to show official providers first

### LINKED ISSUE #178: Provider search for 'keycloak' returns community provider instead of official
## Description
When searching for 'keycloak' in the Terraform Provider Registry via terraform-mcp-server, the top result is a community provider from the user `mrparkers` rather than the official Keycloak provider. This leads to confusion for users who expect official providers to be prioritized.

## Testing plan
1. Search for 'keycloak' using the provider search functionality in terraform-mcp-server.
2. Observe the returned providers and note that the community provider (mrparkers/keycloak) appears instead of the official Keycloak provider.

#### Expected Behavior
The search results should prioritize official (and partner) providers over community ones, especially when searching for well-known software such as Keycloak.

#### Actual Behavior
Currently, the search returns the community provider `mrparkers/keycloak` as the top result, even when an official Keycloak provider exists in the registry.

#### Additional Context
This behavior may cause users to unknowingly use unofficial providers, which could have different support or maintenance guarantees. Please consider updating the provider search logic to prefer official and partner providers when available.

ghost:
Fixes https://github.com/hashicorp/terraform-mcp-server/issues/178

The search_providers tool was returning community providers before official ones in search results. For example, searching for 'keycloak' would show mrparkers/keycloak (community) before keycloak/keycloak (official).

### 1. Enhanced search_providers.go to use Registry v2 provider list API
- Fetches provider tier information (official/partner/community)
- Implements tier-based sorting: official > partner > community  
- Maintains relevance-based sorting within each tier
- Falls back to original behavior if provider list unavailable

### 2. Modified registry.go to support test overrides
- Added SendRegistryCallFn variable for mocking in tests
- Allows unit tests to run without network dependencies

### 3. Added unit tests:
- sort_test.go: Tests tier-based sorting logic
- search_providers_test.go: Integration test with mocked responses
- Ensures prioritization works correctly and prevents regressions

The implementation uses the existing Registry v2 API and maintains backward compatibility. Official and partner providers now appear first in search results, improving discoverability of authoritative providers.

## Testing
All existing tests continue to pass, and new tests verify the prioritization logic works correctly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,8 +19,7 @@ IMPROVEMENTS
 * Updating `mark3labs/mcp-go` and `hashicorp/tfe-go` versions. See [#121](https://github.com/hashicorp/terraform-mcp-server/pull/121)
 * Implementing TLS for the http mode of the MCP server. See [#168](https://github.com/hashicorp/terraform-mcp-server/pull/168)
 * Implemented rate limiting with the MCP server. See [#155](https://github.com/hashicorp/terraform-mcp-server/pull/155)
-* Improved and expanded README with usage instructions for Claude Code. See [#173](https://github.com/hashicorp/terraform-mcp-server/pull/173)
-* Improved and expanded README with usage instructions for Cursor. See [#173](https://github.com/hashicorp/terraform-mcp-server/pull/173)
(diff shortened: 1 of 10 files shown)


gautambaghel (changes requested):
Requesting a test on MCP inspector ^^, with some tests covering keycloak, AWS, Azure & GCP providers

PR #183, opened 2025-09-26
Fix issue with example VS code configuration for the Terraform MCP server

emrousselle:
## PCI review checklist

# Reason for change

A user following the current example to configure VS Code to use the Terraform MCP server would get errors at startup such as "Error in required input: no Terraform token provided, required input: no Terraform token provided".

This is because of an error in the example configuration file when passing parameters to the MCP server.

# Changes made

The example configuration was updated to properly pass the parameters to the MCP server instance, resolving the error messages and allowing the MCP server to connect to the target HCP Terraform or TFE instance.


emrousselle (commented):
While testing the new version of the Terraform MCP server and following the configuration instructions I ran into where the MCP server would not be able to connect to HCP Terraform even when apparently provided with valid parameters (the right fqdn for HCP Terraform and a valid user token).

This PR updates the instructions to show a configuration example that works.

gautambaghel (approved):
LGTM! Feel free to merge!

PR #184, opened 2025-09-28
Fixed Cursor mcp config example

jim-mclean:
The cursor msp.json differs from the vscode standard.  As you'll see in [the docs](https://cursor.com/docs/context/mcp#config-interpolation) the root element should be `mcpServers` not `servers`


gautambaghel (approved):
LGTM!

PR #200, opened 2025-10-19
CI workflow and associated dependencies for Anthropic MCP Registry publishing

leefowlercu:
This PR introduces the following:

1. A new GHA Workflow that runs on Semantic Version Git Tag pushes and uses the `mcp-publisher` CLI to publish this MCP Server's specification to the official Anthropic MCP Server Registry (`https://registry.modelcontextprotocol.io`). See GHA Registry Publishing documentation here: https://github.com/modelcontextprotocol/registry/blob/main/docs/guides/publishing/github-actions.md
2. Introduction of a Docker Label in the `Dockerfile` to assert package provenance. See https://github.com/modelcontextprotocol/registry/blob/main/docs/guides/publishing/publish-server.md#requirements-3
3. A `server.json` file with the specification of this MCP Server's `v0.3.0` version. See `server-json` documentation here: https://github.com/modelcontextprotocol/registry/tree/main/docs/reference/server-json

**NOTE**: I was unable to do a dry-run test of publishing this server using the `server.json` introduced in this PR against a local test registry due to me not understanding some of the expectations of the `mcp-publisher` CLI tool in `--dry-run` mode. I will update this PR when a Dry Run test has been performed and validated. Opening this PR in the meantime to allow Repsitory Maintainers to review the introduced changes.

- Reason for opening this PR: Introduce an automated mechanism for publishing this MCP Server's specification to the Anthropic MCP Server Registry

- How to "undo" this change: Revert the commit that introduces this PR into `main`. Alternatively, simply delete the GHA Workflow script, the `server.json` file, and remove the LABEL introduced into the `Dockerfile`.

- Impact on Security Controls: None. This change does not modify the configuration or application logic of this repository's MCP Server. It simply introduces a new discovery channel.

## PCI review checklist

- [✅] I have documented a clear reason for, and description of, the change I am making.

- [✅] If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request.

- [✅] If applicable, I've documented the impact of any changes to security controls.

### DIFF
--- a/server.json
+++ b/server.json
@@ -2,7 +2,7 @@
   "$schema": "https://static.modelcontextprotocol.io/schemas/2025-10-17/server.schema.json",
   "name": "io.gitub.hashicorp/terraform-mcp-server",
   "title": "Terraform MCP Server",
-  "description": "Terraform MCP Server. Provides tools to interact with TF Public Registry and HCPTF/TFE.",
+  "description": "Generate more accurate Terraform and automate workflows for HCP Terraform and Terraform Enterprise",
   "websiteUrl": "https://github.com/hashicorp/terraform-mcp-server",
   "repository": {
     "url": "https://github.com/hashicorp/terraform-mcp-server",


gautambaghel:
Great work here @leefowlercu - should we only dispatch it in the release branch? that's where we track our releases.

There should also be a smart way to inherit the latest server version I believe, see the [version](https://github.com/hashicorp/terraform-mcp-server/blob/main/version/VERSION) file this is what we use to track the current version being release, if we release a patch version we change this of course before merging into the release branch

server.json
```diff
@@ -0,0 +1,278 @@
+{
+  "$schema": "https://static.modelcontextprotocol.io/schemas/2025-10-17/server.schema.json",
+  "name": "io.modelcontextprotocol.anonymous/terraform-mcp-server",
```
gautambaghel: @leefowlercu - should this be something else apart from "anonymous"?

`io.modelcontextprotocol.anonymous/terraform-mcp-server`

server.json
```diff
@@ -0,0 +1,278 @@
...
+  "title": "Terraform MCP Server",
+  "description": "HashiCorp MCP server for executing Terraform Cloud and Enterprise workflows, resources, and tools.",
```
gautambaghel: Terraform MCP Server for generating more accurate Terraform configuration and executing Terraform Cloud and Enterprise workflows.

server.json
```diff
@@ -0,0 +1,278 @@
...
+  "websiteUrl": "https://github.com/hashicorp/terraform-mcp-server",
+  "repository": {
+    "url": "https://github.com/hashicorp/terraform-mcp-server",
+    "source": "github"
+  },
+  "version": "0.3.0",
```
gautambaghel: If this workflow is manual then the version needs to be an input from the GHA workflow

server.json
```diff
@@ -0,0 +1,278 @@
...
+        {
+          "type": "named",
+          "name": "--rm"
+        },
+        {
+          "type": "named",
+          "name": "-i"
+        },
+        {
+          "type": "positional",
```
gautambaghel, line 265: You still need to pass env vars to the docker run command right? I don't think it automatically takes the env vars declared below unless you do something like

`docker run -it --rm --env-file <(env | grep MCP) hashcorp/terraform-mcp-server`

server.json
leefowlercu: Whoops. That's a placeholder from dry-run `mcp-publisher` testing. Will get this updated.

server.json
leefowlercu: @gautambaghel updated the `server.json` to specify available MCP Server configuration values as Key/Value Environment Variables using `-e` flags to `docker run` (https://docs.docker.com/reference/cli/docker/container/run/#env)

server.json
leefowlercu: Added a Make Target and associated script to update all version specifiers in the `server.json` file with the version from the `version/VERSION` file.

server.json
leefowlercu: Updated the description, but the specification enforces a 100 character maximum for that field which the new description comes in just under.

server.json
leefowlercu: Fixed.

leefowlercu:
Updated `server.json` to describe docker `-e` flags with Key/Value arguments for each of this MCP Servers configuration environment variables.

Constrained the value for the `TRANSPORT_MODE` environment to `stdio` in the Package definition for the `stdio` transport type, and to `streamable-http` for the Package definition for the `streamable-http` transport type.

gautambaghel (approved):
LGTM!

PR #207, opened 2025-10-23
Feature/no-code-modules

gautambaghel:
## PCI review checklist

### DIFF
--- a/pkg/tools/tfe/create_no_code_workspace.go
+++ b/pkg/tools/tfe/create_no_code_workspace.go
@@ -155,7 +155,7 @@ func fetchModuleData(ctx context.Context, tfeClient *tfe.Client, projectID, noCo
 	}
 
 	metadataPath := path.Join("/api/registry/private/v2/modules", registryModule.Namespace, registryModule.Name, registryModule.Provider, "metadata", noCodeModule.VersionPin)
-	metadataData, err := utils.MakeCustomRequestWithDoRaw(ctx, tfeClient, metadataPath, map[string][]string{"organization_name": {noCodeModule.Organization.Name}})
+	metadataData, err := utils.MakeCustomGetRequestRaw(ctx, tfeClient, metadataPath, map[string][]string{"organization_name": {noCodeModule.Organization.Name}})
 	if err != nil {
 		return nil, nil, nil, utils.LogAndReturnError(logger, "making module metadata API request", err)
 	}
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -115,13 +115,9 @@ func GetEnv(key, fallback string) string {
 	return fallback
 }
 
-// This function is used for custom API requests using the TFE client.
-func MakeCustomRequestWithDoRaw(ctx context.Context, client *tfe.Client, path string, additionalQueryParams ...map[string][]string) ([]byte, error) {
-	var queryParams map[string][]string
-	if len(additionalQueryParams) > 0 {
-		queryParams = additionalQueryParams[0]
-	}
-	req, err := client.NewRequestWithAdditionalQueryParams("GET", path, nil, queryParams)
+// This function is used for custom GET requests using the TFE client.
+func MakeCustomGetRequestRaw(ctx context.Context, client *tfe.Client, path string, additionalQueryParams map[string][]string) ([]byte, error) {
+	req, err := client.NewRequestWithAdditionalQueryParams("GET", path, nil, additionalQueryParams)
 	if err != nil {
 		return nil, err
 	}


pkg/tools/dynamic_tool.go
```diff
@@ -243,3 +245,42 @@ func (r *DynamicToolRegistry) createDynamicTFETool(toolName string, toolFactory
 		Handler: wrappedHandler,
 	}
 }
+
+// createDynamicTFEToolWithMCPServer creates a TFE tool with dynamic availability checking that also needs MCPServer
+func (r *DynamicToolRegistry) createDynamicTFEToolWithMCPServer(toolName string, toolFactory func(*log.Logger, *server.MCPServer) server.ServerTool) server.ServerTool {
+	originalTool := toolFactory(r.logger, r.mcpServer)
```
kartheek-hc, line 221: nit: It looks like the this line is the only difference between this method and `createDynamicTFETool` above. In a follow up, can we refactor the rest of the method into a separate method and reuse?

pkg/utils/utils.go
```diff
@@ -111,3 +114,31 @@ func GetEnv(key, fallback string) string {
 	}
 	return fallback
 }
+
+// This function is used for custom API requests using the TFE client.
+func MakeCustomRequestWithDoRaw(ctx context.Context, client *tfe.Client, path string, additionalQueryParams ...map[string][]string) ([]byte, error) {
+	var queryParams map[string][]string
+	if len(additionalQueryParams) > 0 {
+		queryParams = additionalQueryParams[0]
+	}
```
kartheek-hc: nit: Can we make `additionalQueryParams` non-variadic and just pass it directly to `NewRequestWithAdditionalQueryParams`?

pkg/utils/utils.go
```diff
@@ -111,3 +114,31 @@ func GetEnv(key, fallback string) string {
...
+	req, err := client.NewRequestWithAdditionalQueryParams("GET", path, nil, queryParams)
```
kartheek-hc: nit: Since this method is always doing a `GET` request, can we rename the function to `MakeCustomGetRequestWithDoRaw` or simply `MakeCustomGetRequestRaw `?

pkg/tools/tfe/create_no_code_workspace.go
```diff
@@ -0,0 +1,275 @@
...
+			workspaceVariable := &tfe.Variable{
+				Key:      varName,
+				Value:    value,
+				Category: tfe.CategoryTerraform,
+			}
+			variables = append(variables, workspaceVariable)
+		}
+
+	case mcp.ElicitationResponseActionDecline:
+		return nil, utils.LogAndReturnError(logger, "No Code module workspace creation cancelled by user", nil)
```
kartheek-hc: nit: Should we explicitly log `declined`?
```suggestion
	case mcp.ElicitationResponseActionDecline:
		return nil, utils.LogAndReturnError(logger, "No Code module workspace creation declined by user", nil)
```

pkg/tools/tfe/create_no_code_workspace.go
```diff
@@ -0,0 +1,275 @@
...
+	switch result.Action {
+	case mcp.ElicitationResponseActionAccept:
+		// Extract the provided variable values
+		data, ok := result.Content.(map[string]any)
+		if !ok {
+			return nil, utils.LogAndReturnError(logger, "elicitation response content is not a map", fmt.Errorf("expected map[string]any, got %T", result.Content))
+		}
+
+		// Process each provided variable
+		for _, varName := range requiredVars {
```
kartheek-hc: It looks like in the elicitation request above we are passing all variables/properties (required and optional) but we are only reading the ones that are required, is that intentional?

pkg/tools/tfe/create_no_code_workspace.go
```diff
@@ -0,0 +1,275 @@
...
+
+	logger.Infof("Created No Code module workspace: %s", noCodeModuleWorkspace.ID)
+	var buf bytes.Buffer
+	if err := jsonapi.MarshalPayload(&buf, noCodeModuleWorkspace); err != nil {
+		return nil, utils.LogAndReturnError(logger, "marshaling run response", err)
+	}
+
+	return &mcp.CallToolResult{
+		Content: []mcp.Content{
+			mcp.NewTextContent(buf.String()),
```
kartheek-hc: nit: Should we consider trimming the response? The [Workspace](https://pkg.go.dev/github.com/hashicorp/go-tfe#Workspace) struct seems to contain a lot of fields

kartheek-hc (approved):

pkg/tools/tfe/create_no_code_workspace.go
gautambaghel: yes we should, good call

pkg/tools/tfe/create_no_code_workspace.go
gautambaghel: yes, that's the problem overall, I'm hoping we fix it in the next release but I think we do have something for workspace. I'll add that

pkg/tools/tfe/create_no_code_workspace.go
gautambaghel: well the API only returns required variables, but good point maybe I can write this better

kartheek-hc (approved):

PR #215, opened 2025-11-18
enable toolset flag

jaylonmcshan19-x:
## PCI review checklist

This pr Implements support for the `--toolsets` flag.

This PR includes  only the core toolset logic without the tool filtering implementation due to me seeking feedback on:

1. Toolset groupings: Are these logical groupings, or could they be structured differently?
2. Structure of the help message: Is the help message structure clear, or could it be presented in a more user friendly manner
3. Default toolsets: Is this the the right default set of tools?
4. Special keywords: Do the special keywords ('all', 'default') make sense / needed? (I got inspiration on these keywords, from GH MCP server implementation)

### DIFF
--- a/pkg/toolsets/toolsets.go
+++ b/pkg/toolsets/toolsets.go
@@ -23,23 +23,23 @@ type Toolset struct {
 }
 
 var (
-	All_Toolset = Toolset{
+	AllToolset = Toolset{
 		Name:        All,
 		Description: "Special toolset that enables all available toolsets",
 	}
-	Default_Toolset = Toolset{
+	DefaultToolset = Toolset{
 		Name:        Default,
 		Description: "Special toolset that enables the default toolset configuration",
 	}
-	Registry_Toolset = Toolset{
+	RegistryToolset = Toolset{
 		Name:        Registry,
 		Description: "Public Terraform Registry (providers, modules, policies)",
 	}
-	RegistryPrivate_Toolset = Toolset{
+	RegistryPrivateToolset = Toolset{
 		Name:        RegistryPrivate,
 		Description: "Private registry access (TFE/TFC private modules and providers)",
 	}
-	Terraform_Toolset = Toolset{
+	TerraformToolset = Toolset{
 		Name:        Terraform,
 		Description: "HCP Terraform/TFE operations (workspaces, runs, variables, etc.)",
 	}
@@ -47,9 +47,9 @@ var (
 
 func AvailableToolsets() []Toolset {
 	return []Toolset{
-		Registry_Toolset,
-		RegistryPrivate_Toolset,
-		Terraform_Toolset,
+		RegistryToolset,
+		RegistryPrivateToolset,
+		TerraformToolset,
 	}
 }
 
@@ -63,8 +63,8 @@ func GetValidToolsetNames() map[string]bool {
 	for _, ts := range AvailableToolsets() {
 		validNames[ts.Name] = true
 	}
-	validNames[All_Toolset.Name] = true
-	validNames[Default_Toolset.Name] = true
+	validNames[AllToolset.Name] = true
+	validNames[DefaultToolset.Name] = true
 	return validNames
 }
 


pkg/toolsets/mapping.go
```diff
@@ -0,0 +1,84 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package toolsets
+
+// ToolToToolset maps tool names to their toolsets
+var ToolToToolset = map[string]Toolset{
```
jrhouston: As discussed in today's call, we don't want to be so fine grained with the toolsets where we have a toolset for each API resource.

As a starting point here just so we can add support for the toolset flag, I would propose the following toolsets:

- `registry` – for workflows that want to access the public registry 
- `registry-private` – for workflows that access a private registry
- `terraform` – for workflows that need to access TFE/TFC 

In future once we've ironed out personas and their workflows, we can break the `terraform` one into smaller chunks local, admin, readonly-user, etc.

jaylonmcshan19-x: Sounds good, will fix in the next commit!

jaylonmcshan19-x:
@gautambaghel Yup, just waiting on a review of my implementation and will add tests tomorrow! Once approved will merge 👍

pkg/toolsets/mapping.go
```diff
@@ -0,0 +1,71 @@
...
```
jrhouston: I think maintaining a map like this is OK for now, but I suspect in future we'll want to make this easier to maintain by having the tool initialization and assignment happening in the same place, i.e. in [tools.go](pkg/tools/tools.go).

The GitHub MCP server project does something similar here: https://github.com/github/github-mcp-server/blob/ada4bc05a9c7c8246bd0651386a558802e23dec0/pkg/github/tools.go#L163-L379

pkg/toolsets/toolsets.go
```diff
@@ -0,0 +1,162 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package toolsets
+
+import "strings"
+
+// Toolset represents a group of related tools
+type Toolset string
```
jrhouston: I'd perhaps call this `ToolsetName` because that's what it actually is.

pkg/toolsets/toolsets.go
```diff
@@ -0,0 +1,162 @@
...
+	ToolsetRegistry        Toolset = "registry"         // Public Terraform Registry
+	ToolsetRegistryPrivate Toolset = "registry-private" // Private registry (TFE/TFC)
+	ToolsetTerraform       Toolset = "terraform"        // TFE/TFC operations
+
+	// Special toolsets
+	ToolsetAll     Toolset = "all"
+	ToolsetDefault Toolset = "default"
+)
+
+type ToolsetMetadata struct {
```
jrhouston: This one I might just call `Toolset`. `Metadata` such an overloaded concept already throughout our projects.

pkg/toolsets/toolsets.go
```diff
@@ -0,0 +1,162 @@
...
+	ID          string
```
jrhouston: Above you defined a type for this, but here you're using string and calling it `ID`. If you rename you `Toolset` type to `ToolsetName` this line becomes 

```golang
Name ToolsetName
```

pkg/toolsets/toolsets.go
```diff
@@ -0,0 +1,162 @@
...
+			if !validIDs[trimmed] {
+				invalid = append(invalid, trimmed)
+			}
+		}
+	}
+
+	return result, invalid
+}
+
+func ExpandDefaultToolset(toolsets []string) []string {
```
jrhouston, line 94: Similarly, here (and below in `ContainsToolset`) you have used `toolsets []string`, but we define a specific type `Toolset` for this above. To be consistent I would either not define the type – or use it here. Since we're not actually adding any receivers to the `Toolset` type I'd be inclined to just use `string` everywhere.

jrhouston (changes requested):

pkg/toolsets/toolsets.go
```diff
@@ -0,0 +1,155 @@
...
+	Name        string
+	Description string
+}
+
+var (
+	All_Toolset = Toolset{
+		Name:        All,
+		Description: "Special toolset that enables all available toolsets",
+	}
+	Default_Toolset = Toolset{
```
jrhouston: Nit: we're not supposed to use underscores in variable names.

jrhouston (changes requested):
One small thing.

PR #265, opened 2026-01-26
Add credentials.tfrc.json support for TFE authentication (#228)

### LINKED ISSUE #228: Alternate way to pass TFE/TFC variables to MCP service
#### Use-cases
In the current implementation, sensitive flags (e.g. TFE_TOKEN) are read from an environment variable. This is ok for the MCP running inside a container since the variables are passed during the container launch but can be problematic for customers not permitted to run docker locally as the token would need to be declared as an env variable on the host machine. This could lead to token leak

#### Attempted Solutions
Attempted to leverage flags in the off chance they were already present. This failed.

#### Proposal
1) Add flags for the env variables so they can be passed directly to the server during initialization. i.e. --tfe_server and --tfe_token
2) Configure the MCP service to look for the normal credentials.tfrc.json and read the token from there.

vravind1:
Fixes #228                                                                                               
                                                                                                           
  - Adds support for reading TFE tokens from `~/.terraform.d/credentials.tfrc.json`                        
  - Users who have run `terraform login` can now use the MCP server without setting `TFE_TOKEN` environment
   variable                                                                                                
  - Maintains backward compatibility - env var still takes precedence when set                             
                                                                                                           
  **New credential resolution chain:**                                                                     
  HTTP Header → Environment Variable → credentials.tfrc.json → Error                                       
                                                                                                           
  ## Changes                                                                                               
                                                                                                           
  | File | Description |                                                                                   
  |------|-------------|                                                                                   
  | `pkg/client/credentials.go` | New file - reads and parses credentials.tfrc.json |                      
  | `pkg/client/credentials_test.go` | Unit tests for credential reading |                                 
  | `pkg/client/tfe_client.go` | Added fallback to credentials file (3 lines) |                            
                                                                                                           
  ## Test plan                                                                                             
                                                                                                           
  - [x] Unit tests pass (`go test ./pkg/client/...`)                                                       
  - [x] All existing tests pass (`go test ./...`)                                                          
  - [x] Manual testing with Claude Desktop - successfully queried HCP Terraform workspaces using token from
   credentials file


pkg/client/credentials_test.go
```diff
@@ -0,0 +1,108 @@
...
+func TestExtractHostname(t *testing.T) {
+	tests := []struct {
+		name     string
+		address  string
+		expected string
+	}{
+		{"standard https", "https://app.terraform.io", "app.terraform.io"},
+		{"with port", "https://tfe.example.com:8443", "tfe.example.com"},
+		{"with path", "https://app.terraform.io/api/v2", "app.terraform.io"},
+		{"http scheme", "http://localhost:8080", "localhost"},
```
jrhouston: TBH – since extractHostname is a private function that just adds some sugar on top of `url.Parse`, I don't see much value in these tests. They are just re-testing the `url.Parse()` function which already has a comprehensive suite of tests that look exactly like this. The logic we've introduced here is returning `""` for empty input or invalid URL.

pkg/client/credentials.go
```diff
@@ -0,0 +1,65 @@
...
+	Credentials map[string]credentialEntry `json:"credentials"`
+}
+
+type credentialEntry struct {
+	Token string `json:"token"`
+}
+
+// ReadCredentialsFile reads the Terraform CLI credentials file and returns
+// the token for the specified hostname. Returns empty string if not found.
+func ReadCredentialsFile(hostname string) string {
```
jrhouston: This works but the function fails silently, and there's a few real failure modes we'd probably want to at least warn the user about in the logs:

1. The path / file simply doesn't exist
2. We tried to read the file but don't have permissions to do so. 
3. We read the file successfully but couldn't parse it.
4. We read and parse the file successfully but the hostname we've configured doesn't exist in the file. 

Right now if I try to use the credentials file but misconfigure it in some way, the server boots up OK but now I have a task of trying to figure out why it doesn't like my credentials file.

jrhouston (changes requested):
Hey Vishnu 😄 

Looks good, just a couple of test/error handling comments.

pkg/client/credentials_test.go
vravind1: I've trimmed the test cases down to only the edge cases

pkg/client/credentials.go
rmcolbert: @jrhouston, can you review again after additional updates were made to see if your concerns have been addressed?

Shweta-Deshpande:
Taking over - https://github.com/hashicorp/terraform-mcp-server/pull/333

PR #344, opened 2026-04-25
Add module examples and submodules tools

### LINKED ISSUE #307: Fetch module examples and submodules independently from `get_module_details`
#### Use-cases

Some modules can have many examples or submodules. I want the MCP to selectively surface the examples and submodules. Currently, `get_module_details` takes a module ID as input and returns all the module examples in a single section. This may not always be a good idea, as a module may be multi-purpose and not all examples are helpful for the current task. Always including all examples can consume a lot of tokens for not much benefit.

`get_module_details` also doesn't surface submodules at all, which may be a separate issue. But the idea is also the same. I don't want to get all the submodules at once; the agent should be able to learn about a specific submodule.

#### Attempted Solutions

Called the Registry API directly and checked the result.

#### Proposal

Two possible solutions:
- Create new tools to enable example and submodule lookup. Under the hood, they just use the same REST API but extract different pieces of information.
- Overload the existing `get_module_details` tool to be more customizable. Depending on the module ID suffix, return the additional information. For example:
  - `namespace/name/provider/version`: Return the root module details + a list of example names and submodule names (and any lightweight metadata for examples and submodules) + instructions on how to query examples/submodules.
  - `namespace/name/provider/version/examples`: Return all example details of that module.
  - `namespace/name/provider/version/examples/example_name`: Return a specific example.
  - `namespace/name/provider/version/submodules`: Return all submodule details of that module.
  - `namespace/name/provider/version/submodules/submodule_name`: Return a specific submodule's details.

ayhammouda:
## Summary
- add get_module_examples and get_module_submodules registry tools
- keep get_module_details focused on root module details plus example/submodule indexes
- add unit and Docker e2e coverage for listing and selecting module examples/submodules

Closes #307

## Validation
- go test ./pkg/tools/registry ./pkg/tools ./pkg/toolsets
- go vet ./pkg/tools/registry ./pkg/tools ./pkg/toolsets
- go test ./cmd/terraform-mcp-server ./pkg/client ./pkg/resources ./pkg/tools ./pkg/tools/registry ./pkg/tools/tfe ./pkg/toolsets ./pkg/utils ./version
- go vet ./cmd/terraform-mcp-server ./pkg/client ./pkg/resources ./pkg/tools ./pkg/tools/registry ./pkg/tools/tfe ./pkg/toolsets ./pkg/utils ./version
- make test-e2e

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ IMPROVEMENTS
 FEATURES
 
 * [New Tool] `get_sentinel_mock` Export and download Sentinel mock bundle data for a Terraform plan
+* [New Tool] `get_module_examples` and `get_module_submodules` retrieve Terraform module examples and submodules separately from root module details [344](https://github.com/hashicorp/terraform-mcp-server/pull/344)
 
 IMPROVEMENTS
 
--- a/cmd/terraform-mcp-server/instructions.md
+++ b/cmd/terraform-mcp-server/instructions.md
@@ -20,6 +20,8 @@ The Terraform MCP server provides tools for generating better Terraform code thr
   - `get_provider_capabilities` shows what types of resources, data sources, functions, and guides are available
   
 - **Module Discovery**: `get_latest_module_version` (if unavailable in code) → `search_modules` → `get_module_details`
+  - `get_module_details` returns root module requirements plus lightweight example and submodule indexes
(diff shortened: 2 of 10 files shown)


ayhammouda:
Hi! Quick follow-up on this PR since the checks are passing now.
Happy to adjust anything if needed. Thanks!

PR #337, opened 2026-04-21
disallow TFE_ADDRESS header when TFE_TOKEN is set via env var

jaylonmcshan19-x:
## PCI review checklist

This pr fixes a security vulnerability where a  user could redirect API requests to a server they control to harvest the `Authorization` header containing the server's `TFE_TOKEN`.

- [ ] If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request.

- [ ] If applicable, I've documented the impact of any changes to security controls.

  Examples of changes to security controls include using new access control methods, adding or removing logging pipelines, etc.

PR #348, opened 2026-05-01
Bootstrap the server using the official mcp go-sdk

Shweta-Deshpande:
## PCI review checklist

### DIFF
--- a/pkg/mcp-official/server.go
+++ b/pkg/mcp-official/server.go
@@ -1,0 +1,243 @@
+package mcpofficial
+
+import (
+	"context"
+	"crypto/tls"
+	"fmt"
+	"net/http"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/hashicorp/go-cleanhttp"
+	"github.com/hashicorp/go-retryablehttp"
+	"github.com/hashicorp/go-tfe"
+	"github.com/hashicorp/terraform-mcp-server/pkg/utils"
+	"github.com/hashicorp/terraform-mcp-server/version"
+	"github.com/modelcontextprotocol/go-sdk/mcp"
+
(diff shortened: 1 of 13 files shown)


jaylonmcshan19-x (dismissed):
LGTM. PR #337 adds a security check that blocks TFE_ADDRESS headers when TFE_TOKEN comes from env var. Once that merges, the new helper will need the same logic I believe

pkg/mcp-official/server.go
```diff
@@ -0,0 +1,143 @@
...
+
+		// Optional fields (will be empty strings if not provided)
+		ProjectID    string `json:"project_id,omitempty" jsonschema:"Filter by project ID"`
+		SearchQuery  string `json:"search_query,omitempty" jsonschema:"Search term"`
+		Tags         string `json:"tags,omitempty" jsonschema:"Comma-separated tags"`
+		ExcludeTags  string `json:"exclude_tags,omitempty" jsonschema:"Tags to exclude"`
+		WildcardName string `json:"wildcard_name,omitempty" jsonschema:"Wildcard pattern"`
+	}
+
+	mcp.AddTool(svr, &mcp.Tool{
```
jaylonmcshan19-x, line 68: Quick question! what's the plan for testing the new SDK tools? I believe the existing tests check tool creation via the returned ServerTool struct, but with mcp.AddTool() being inline, I think we might need a different approach.

Shweta-Deshpande: Yeah will look into it and create a new ticket and PR for that.

Shweta-Deshpande: Correction: migration of existing e2e tests will also be done via this PR. Thanks @jaylonmcshan19-x

PR #355, opened 2026-05-07
Add new metric to capture client type

Shweta-Deshpande:
As part of better understanding the usage of our mcp server we need to be able to answer:
Where are customers calling our mcp server from? Claude, Bedrock etc..
Add a metric which can capture the client type.
We can do this by creating a hook for the Initialize lifecycle step and grabbing the clientInfo see spec: 
[Lifecycle - Model Context Protocol](https://modelcontextprotocol.io/specification/2025-11-25/basic/lifecycle#initialization)

## PCI review checklist

### DIFF
--- a/cmd/terraform-mcp-server/main.go
+++ b/cmd/terraform-mcp-server/main.go
@@ -34,6 +34,7 @@ import (
 
 //go:embed instructions.md
 var instructions string
+var sessionClientInfo sync.Map // map[string]client.ClientInfo
 
 func runHTTPServer(logger *log.Logger, host string, port string, endpointPath string, heartbeatInterval time.Duration, enabledToolsets []string, metricsConfig client.MetricsConfig) error {
 	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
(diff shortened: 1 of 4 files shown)


cmd/terraform-mcp-server/main.go
```diff
@@ -78,6 +78,13 @@ func attachMetricsHooks(hooks *server.Hooks, metricsConfig client.MetricsConfig,
 	if !metricsConfig.Enabled {
 		return
 	}
+	hooks.AddAfterInitialize(func(ctx context.Context, id any, message *mcp.InitializeRequest, result *mcp.InitializeResult) {
+		if message != nil && message.Params.ClientInfo.Name != "" {
+			clientName := message.Params.ClientInfo.Name
+			clientVersion := message.Params.ClientInfo.Version
+			client.RecordClientType(ctx, clientName, clientVersion, metricsConfig, logger)
```
jrhouston: This ClientInfo struct has a few fields: https://github.com/mark3labs/mcp-go/blob/v0.52.0/mcp/types.go#L610

Perhaps we should capture Title and Description too, if it's there?

pkg/client/metrics.go
```diff
@@ -103,3 +104,19 @@ func RecordToolCall(ctx context.Context, startTime time.Time, toolErr bool, id a
 		logger.Errorf("Recorded error for tool %s", message.Params.Name)
 	}
 }
+
+// RecordClientType records the type and version of the client making the tool call (e.g., CLI, VSCode, Web, etc.)
+func RecordClientType(ctx context.Context, clientName string, clientVersion string, config MetricsConfig, logger *log.Logger) {
+	logger.Infof("Recording client type for client: %s version: %s", clientName, clientVersion)
+	if !config.Enabled || config.ClientTypeCounter == nil {
+		logger.Errorf("DEBUG: Either metrics are not enabled or ClientTypeCounter is NIL! Initialization failed.")
```
jrhouston: I know this is copypasta from above, but since this is just a debug statement, it should be `logger.Debugf` and we don't need to write "DEBUG" in the message.

jrhouston (changes requested):
Couple small things

pkg/client/metrics.go
Shweta-Deshpande: Done

cmd/terraform-mcp-server/main.go
Shweta-Deshpande: Done

cmd/terraform-mcp-server/main.go
```diff
@@ -78,10 +79,51 @@ func attachMetricsHooks(hooks *server.Hooks, metricsConfig client.MetricsConfig,
...
+				return
+			}
+			ci := client.ClientInfo{
+				Name:        message.Params.ClientInfo.Name,
+				Version:     message.Params.ClientInfo.Version,
+				Title:       message.Params.ClientInfo.Title,
+				Description: message.Params.ClientInfo.Description,
+			}
+			// Record the client info in the session first so we can reuse it in the BeforeToolCall hook
+			sessionClientInfo.Store(session.SessionID(), ci)
```
jrhouston, line 98: Because we're storing this so it can be used in the Tool call, should we after a hook that removes this from the map when the connection is closed?

jrhouston (changes requested):

cmd/terraform-mcp-server/main.go
Shweta-Deshpande: Done

jrhouston (approved):

PR #356, opened 2026-05-08
Run runtime images as non-root for Kubernetes compatibility

sammcgeown:
The runtime container images were running as root by default (scratch image + no explicit user). In Kubernetes clusters with restricted Pod Security settings, this can fail admission or violate security policy when runAsNonRoot is required.

**Changes Made**
Added an explicit non-root runtime user to the dev image stage:
- Added the same explicit non-root runtime user to the release image stage:
- Chosen user/group is numeric 65532:65532, which works in scratch-based images without requiring passwd/group files.

This change ensures containers do not run as UID 0 by default. It improves compatibility with Kubernetes securityContext settings such as `runAsNonRoot: true`. Keeps runtime behaviour unchanged aside from process identity (entrypoints and binaries are unchanged).

**Scope/Impact:**
- Runtime security hardening only.
- No API, protocol, or functional behaviour changes.

## PCI review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+IMPROVEMENTS
+
+* Run as a non-root user for Kubernetes compatibility. [356] https://github.com/hashicorp/terraform-mcp-server/pull/356
+
 # 0.5.2
 
 FEATURES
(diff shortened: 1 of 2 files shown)


gautambaghel:
and openshift ;)

jrhouston (approved):

PR #358, opened 2026-05-08
Use configurable container runtime in e2e tests instead of hardcoded Docker commands

sammcgeown:
The e2e test flow hardcoded `docker` in test helper commands, even though the Makefile already supports a configurable container runtime via `DOCKER` variable.

That caused `make test-e2e` to fail in environments using Podman or another Docker-compatible runtime, because the Go tests ignored the configured runtime and invoked `docker` directly.

## Changes Made
- Added a shared helper in the e2e test package to resolve the container runtime from the DOCKER environment variable.
- Kept a docker fallback so existing Docker-based workflows continue to work unchanged.
- Replaced hardcoded docker invocations in e2e test helpers with the resolved runtime.
- Updated the test-e2e Makefile target to export DOCKER=$(DOCKER) into the Go test process so the tests use the same runtime as the Makefile.

## Why This Fix
- Aligns the Go e2e tests with the existing Makefile configuration.
- Allows make test-e2e to work with Podman without requiring local aliases or manual edits.
- Preserves backward compatibility for Docker users.

## Scope
Test infrastructure only.
No changes to server runtime behaviour or production container images.

## PCI review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+
+* Use the configured container runtime for e2e tests instead of hardcoded Docker commands, enabling Podman-based test runs [358](https://github.com/hashicorp/terraform-mcp-server/pull/358)
+
 # 0.5.2
 
 FEATURES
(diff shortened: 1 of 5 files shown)


gautambaghel:
this is cool, should help us support podman :)

PR #362, opened 2026-05-13
ci(release-checks): declare contents: read

arpitjain099:
The version-consistency workflow currently inherits the default `GITHUB_TOKEN` scope. It only runs `scripts/compare-versions.sh` against a checkout, so `contents: read` is the correct minimum.

Matches the top-level pattern already in `unit_test.yml`, `changelog.yml`, `publish-registry.yml`, `e2e_test.yml`. YAML validated locally.

### DIFF
--- a/.github/workflows/release-checks.yml
+++ b/.github/workflows/release-checks.yml
@@ -2,11 +2,14 @@ name: Version Consistency Check
 
 on:
   push:
-    branches: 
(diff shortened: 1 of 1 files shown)


arpitjain099:
Closing this for now. Thanks for the review consideration; may revisit in the future if needed.

PR #371, opened 2026-05-25
Add force_unlock_workspace tool

### LINKED ISSUE #372: Add force_unlock_workspace MCP tool
#### Use-cases

When a Terraform workspace lock gets stuck (e.g. a run crashes or is interrupted and the lock isn't released, or the locking session times out without releasing), there is no way today to recover from inside an AI assistant context.

Operators have to leave their MCP-driven flow and either:

- Click "Force unlock" in the HCP Terraform/TFE UI, or
- Hand-roll `POST /api/v2/workspaces/:id/actions/force-unlock` via curl with their TFE token.

Both options break the assistant loop and the second is the kind of action operators tend to get wrong under pressure (correct URL, correct headers, correct method on the right workspace ID).

#### Attempted Solutions

- `delete_workspace_safely` doesn't apply — the workspace is healthy, just locked.
- `action_run` (cancel/discard) only helps if the lock is held by a specific run that is still active. A stuck or orphaned lock can't be cleared this way.
- The go-tfe client already exposes the operation as `client.Workspaces.ForceUnlock(ctx, workspaceID)`, so it's a small wrap of an existing primitive.

We're already running a bespoke Lambda + Bedrock AgentCore gateway target internally at Aircall to expose this single operation. It would be much simpler to retire that and rely on a first-class tool here.

#### Proposal

Add a `force_unlock_workspace` MCP tool patterned on `delete_workspace_safely`:

- Input: `workspace_id` (required, e.g. `ws-abc123…`).
- Behaviour: call `tfeClient.Workspaces.ForceUnlock(ctx, workspaceID)`.
- Annotations: `ReadOnlyHint=false`, `DestructiveHint=true`, `OpenWorldHint=true`.
- Toolset: `Terraform`.
- Gated behind `ENABLE_TF_OPERATIONS=true`, same opt-in flag as `delete_workspace_safely` and `action_run`. Default off.
- TFE-side authorisation is enforced by the existing `can-force-unlock` permission on the workspace, so the tool inherits that.
- Description should explicitly warn that force-unlocking while a run is still active can leave state inconsistent, and that the safe fallback is to cancel/discard the run via `action_run` first.

Implementation is in PR #371.

Open questions for maintainers (happy to adjust in review):

1. **Input shape.** I went with `workspace_id` to match `delete_workspace_safely` and the underlying go-tfe API. `get_workspace_details` instead takes `terraform_org_name` + `workspace_name`. If maintainers would rather accept org+name here (the recovery use case often starts from a name, not an ID), I'll change it.
2. **Gating.** Reusing `ENABLE_TF_OPERATIONS` keeps the matrix simple. If a finer-grained flag is preferred (e.g. `ENABLE_TF_FORCE_UNLOCK`), happy to add it.

sebin:
Closes #372.

## Summary
- New TFE tool `force_unlock_workspace` that force-unlocks a Terraform workspace by ID via go-tfe `Workspaces.ForceUnlock`.
- Intended as a recovery action for stuck workspace locks (e.g. after an interrupted run).
- Annotated as destructive (`ReadOnlyHint=false`, `DestructiveHint=true`, `OpenWorldHint=true`) and gated behind `ENABLE_TF_OPERATIONS`, matching `delete_workspace_safely` and `action_run`.

## Why
The `force-unlock` workspace action is a recovery path operators need but it isn't exposed by the MCP server today. Engineers hitting a stuck workspace lock have to leave their AI assistant context and call the TFE API by hand. Adding it here closes that gap while keeping the operation gated behind the same opt-in flag as the other destructive workspace tools. See #372 for the full feature-request rationale and open questions.

## Changes
- `pkg/tools/tfe/force_unlock_workspace.go` — tool definition + handler, follows the `delete_workspace_safely` pattern.
- `pkg/tools/tfe/force_unlock_workspace_test.go` — unit tests for tool creation, annotations, parameter validation, ID format.
- `pkg/tools/dynamic_tool.go` — registers the tool when `ENABLE_TF_OPERATIONS=true` and the toolset is enabled.
- `pkg/toolsets/mapping.go` — maps `force_unlock_workspace` to the `Terraform` toolset.
- `cmd/terraform-mcp-server/instructions.md` — documents the new tool under Workspace Management.
- `CHANGELOG.md` — entry under 0.5.3 FEATURES.

## Test plan
- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test ./...` — 583 tests pass (10 packages)
- [x] `go test ./pkg/tools/tfe/ -run TestForceUnlock -v` — 12 new assertions pass

## PCI review checklist

- [x] If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request. (Pure additive: revert this PR to remove the tool. No data/schema migrations.)
- [x] If applicable, I've documented the impact of any changes to security controls. (No change to existing controls. The new tool is destructive and is therefore gated behind `ENABLE_TF_OPERATIONS`, the same opt-in flag already used by `delete_workspace_safely` and `action_run`. It additionally requires a valid `TFE_TOKEN` whose principal must have `can-force-unlock` permission on the target workspace; the underlying TFE API enforces that.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 FEATURES
 
+* [New Tool] `force_unlock_workspace` Force-unlocks a Terraform workspace by ID via the go-tfe `Workspaces.ForceUnlock` API; destructive, gated behind `ENABLE_TF_OPERATIONS` [371](https://github.com/hashicorp/terraform-mcp-server/pull/371)
 * [New Tool] `get_sentinel_mock` Export and download Sentinel mock bundle data for a Terraform plan
 
 IMPROVEMENTS
--- a/cmd/terraform-mcp-server/instructions.md
+++ b/cmd/terraform-mcp-server/instructions.md
@@ -34,8 +34,9 @@ The Terraform MCP server provides tools for generating better Terraform code thr
 
 ### Workspace Management
 - **Discovery**: `search_workspaces` (empty query returns all) → `get_workspace_details`
-- **Operations**: `create_workspace`, `update_workspace`, `delete_workspace_safely`
+- **Operations**: `create_workspace`, `update_workspace`, `delete_workspace_safely`, `force_unlock_workspace`
 - `delete_workspace_safely` only works if workspace has no managed resources
+- `force_unlock_workspace` recovers a workspace whose lock is stuck; destructive — use only after the responsible run has been completed or cancelled
 
 ### Run Execution
 - **Discovery**: `search_run` (empty query returns all) → `get_run_details` (supports json output)
--- a/pkg/tools/dynamic_tool.go
+++ b/pkg/tools/dynamic_tool.go
@@ -143,6 +143,12 @@ func (r *DynamicToolRegistry) registerTFETools() {
 		r.mcpServer.AddTool(tool.Tool, tool.Handler)
 	}
 
+	// Only register force_unlock_workspace if TF operations are enabled AND toolset is enabled
+	if isTerraformOperationsEnabled() && toolsets.IsToolEnabled("force_unlock_workspace", r.enabledToolsets) {
+		tool := r.createDynamicTFETool("force_unlock_workspace", tfeTools.ForceUnlockWorkspace)
+		r.mcpServer.AddTool(tool.Tool, tool.Handler)
+	}
+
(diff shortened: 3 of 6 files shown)


jrhouston:
Closing in favour of #387

PR #377, opened 2026-05-28
Fix: --tools/--toolsets ignored in streamable-HTTP env-var mode (#376)

### LINKED ISSUE #376: --tools and --toolsets flags silently ignored in streamable-HTTP env-var mode
## Summary

`--tools` and `--toolsets` CLI flags are silently ignored when the server is started in streamable-HTTP mode via the `TRANSPORT_MODE` env var. The flags always fall back to their declared defaults (`--toolsets="all"`, `--tools=""`), regardless of what is passed on the command line. The flags work as expected with the explicit `terraform-mcp-server streamable-http …` subcommand invocation — only the env-var-driven shortcut is affected.

This is a regression for anyone running the server in containers/Kubernetes/Bedrock AgentCore, where the convention is to set transport via env vars and tool filtering via flags. The server then exposes every tool (including destructive HCP Terraform writes) when only registry/read tools were requested.

## Reproduction

Tested against `hashicorp/terraform-mcp-server:0.5.2` (also present on `main`, commit 5ca056e):

```bash
docker run -d --rm \
  -e TRANSPORT_MODE=streamable-http \
  -e TRANSPORT_HOST=0.0.0.0 \
  -e TRANSPORT_PORT=8080 \
  -e MCP_SESSION_MODE=stateless \
  -e TFE_TOKEN=dummy \
  -p 18080:8080 \
  hashicorp/terraform-mcp-server:0.5.2 \
  --tools=list_workspaces,get_workspace_details

# Verify the binary received --tools correctly:
docker exec <container> ps -ef
#   PID   USER     COMMAND
#     1   root     terraform-mcp-server streamable-http --tools=list_workspaces,get_workspace_details

# tools/list returns the full 43-tool catalog instead of the requested 2:
curl -s -X POST http://localhost:18080/mcp \
  -H 'Content-Type: application/json' \
  -H 'Accept: application/json, text/event-stream' \
  -d '{"jsonrpc":"2.0","id":1,"method":"tools/list"}' \
  | grep -oE '"name":"[a-zA-Z_]+"' | sort -u | wc -l
# → 43 (including attach_policy_set_to_workspaces, create_workspace, create_run, …)

# Startup log:
#   level=info msg="Starting in Streamable HTTP mode based on environment configuration"
#   level=info msg="Enabled toolsets: [all]"
# (no "Enabled individual tools" line — confirming the flag was never parsed)
```

The same invocation **without** `TRANSPORT_MODE` (taking the cobra subcommand path) honours `--tools` correctly.

## Root cause

`cmd/terraform-mcp-server/main.go` (current `main`, lines 286–301):

```go
if shouldUseStreamableHTTPMode() {
    logger.Info("Starting in Streamable HTTP mode based on environment configuration")
    …
    enabledToolsets := getToolsetsFromCmd(rootCmd, logger)
    …
    if err := runHTTPServer(…); err != nil { … }
    return
}

// Fall back to normal CLI behavior
if err := rootCmd.Execute(); err != nil { … }
```

When `shouldUseStreamableHTTPMode()` returns true (because `TRANSPORT_MODE`/`TRANSPORT_HOST`/`TRANSPORT_PORT`/`MCP_ENDPOINT` is set), the env-var-driven fast path runs **before** `rootCmd.Execute()` is called. cobra's flag parser is therefore never invoked, so:

* `rootCmd.PersistentFlags().GetString("toolsets")` returns the declared default `"all"`.
* `rootCmd.PersistentFlags().GetString("tools")` returns `""`.

`getToolsetsFromCmd(rootCmd, logger)` then takes the `--toolsets="all"` branch unconditionally and `parseIndividualTools` is never reached.

The explicit `streamable-http` subcommand path works because cobra's `Execute()` parses flags before calling the subcommand's `Run`.

## Suggested fix

Move the env-var-driven HTTP-mode detection inside `runDefaultCommand` (the default Run of `rootCmd`), so cobra's `Execute()` always runs and parses persistent flags first. `main()` collapses to:

```go
func main() {
    if err := rootCmd.Execute(); err != nil {
        fmt.Println(err)
        os.Exit(1)
    }
}
```

…and the streamable-HTTP-via-env branch moves to `runDefaultCommand`, *after* `getToolsetsFromCmd(cmd, logger)`. Happy to send a PR.

## Environment

* `terraform-mcp-server` v0.5.2 (also reproducible against `main` at 5ca056e)
* `TRANSPORT_MODE=streamable-http`
* Both `hashicorp/terraform-mcp-server:0.5.2` Docker image and binaries built from source

sebin:
## Summary

Fixes #376.

When the server was launched via env vars (`TRANSPORT_MODE` / `TRANSPORT_HOST` / `TRANSPORT_PORT` / `MCP_ENDPOINT`) instead of the explicit `streamable-http` subcommand, `main()` called `runHTTPServer` directly without ever calling `rootCmd.Execute()`. cobra therefore never parsed `os.Args[1:]`, so both `--tools` and `--toolsets` always fell back to their declared defaults (`""` and `"all"`), regardless of what the operator passed on the command line. The HCP-Terraform toolset was silently enabled in full whenever `TFE_TOKEN` was present, even when the operator asked for a smaller set.

## Change

* Move the env-var-driven HTTP-mode detection inside `runDefaultCommand` (the default `Run` of `rootCmd`), so cobra's `Execute()` always runs first and the `--tools` / `--toolsets` persistent flags are populated before `getToolsetsFromCmd` reads them.
* Collapse `main()` to a plain `rootCmd.Execute()`.
* The explicit \`streamable-http\` subcommand path is unchanged.
* No exported API changes; no behavior change for stdio mode.

Diff is +18 / -18.

## Verification

\`\`\`bash
TRANSPORT_MODE=streamable-http MCP_SESSION_MODE=stateless TFE_TOKEN=x \\
  ./terraform-mcp-server --tools=list_workspaces,get_workspace_details &

curl -s -X POST http://localhost:8080/mcp \\
  -H 'Content-Type: application/json' \\
  -H 'Accept: application/json, text/event-stream' \\
  -d '{\"jsonrpc\":\"2.0\",\"id\":1,\"method\":\"tools/list\"}' \\
  | grep -oE '\"name\":\"[a-zA-Z_]+\"' | sort -u | wc -l
\`\`\`

* Before: **43** (full catalog).
* After: **2** (only the requested tools).

Existing test suite passes:

\`\`\`
$ go test ./...
ok      ...  571 passed in 10 packages
\`\`\`

## Test plan

* [x] All existing tests pass (\`go test ./...\`).
* [x] Manual repro from #376 returns the requested 2 tools instead of 43.
* [x] \`stdio\` subcommand unchanged — still honours \`--tools\` / \`--toolsets\`.
* [x] Explicit \`streamable-http\` subcommand unchanged — still honours \`--tools\` / \`--toolsets\`.

## Notes for reviewers

I did not add a new test because \`runDefaultCommand\` blocks on the server until SIGINT and the existing tests don't have a harness for that path. Happy to add one (e.g. extracting the toolset-selection block into a helper that takes a stub for the server-run function) if maintainers want it.


sebin:
Heads-up on a side effect of the structural change that I should have called out in the PR body — the four-line logger setup at the top of the old `main()` was removed:

```go
logFile, _ := rootCmd.PersistentFlags().GetString("log-file")
logLevel := getLogLevel(rootCmd)
logFormat := getLogFormat(rootCmd)
logger, err := initLogger(logFile, logLevel, logFormat)
```

These reads ran before `rootCmd.Execute()` and therefore had the same defect as the `--tools` / `--toolsets` reads — they always returned the declared defaults (`""`, `"info"`, `"text"`) regardless of what the operator passed as `--log-file=…` / `--log-level=debug` / `--log-format=json`. The resulting `logger` was only consumed by the env-var-driven HTTP branch; every other code path (`stdioCmd`, `streamableHTTPCmd`, `runDefaultCommand`) constructs its own logger inside its `Run` after cobra has parsed flags.

After the fix, the env-var-driven HTTP branch lives inside `runDefaultCommand`, which already calls `initLogger(…)` with correctly-parsed flag values. The top-of-`main()` block became both redundant and incorrect, so it was removed. CLI log flags now actually take effect in the env-var-driven path — a small bonus fix from the same structural change.

jaylonmcshan19-x:
Hey @sebin Thanks for finding the issue and providing the fix :)
One final ask: Can you include a regression test? Given the bug silently exposed the whole catalog when someone asked for a subset, I'd be nice to have something locking the behavior down.

PR #380, opened 2026-06-02
Fix: --tools alone falsely flagged as conflicting with --toolsets

### LINKED ISSUE #379: --tools alone exits with "Cannot use both --tools and --toolsets flags together"
## Summary

Passing only `--tools` (without `--toolsets`) on the command line causes the server to exit immediately with:

```
Cannot use both --tools and --toolsets flags together
```

## Repro

```
$ go build -o ./terraform-mcp-server ./cmd/terraform-mcp-server/
$ ./terraform-mcp-server --tools=search_providers,get_provider_details
time="..." level=fatal msg="Cannot use both --tools and --toolsets flags together"
$ echo $?
1
```

The same Fatal fires whether the transport is `stdio`, the `streamable-http` subcommand, or the env-var-driven HTTP path (`TRANSPORT_MODE=streamable-http ./terraform-mcp-server --tools=...`).

## Root cause

`getToolsetsFromCmd` in `cmd/terraform-mcp-server/main.go` reads the `--toolsets` value with `GetString` and rejects anything that is non-empty and not `"default"`:

```go
toolsetsFlag, _ := cmd.Flags().GetString("toolsets")
if toolsetsFlag == "" {
    toolsetsFlag, _ = cmd.Root().PersistentFlags().GetString("toolsets")
}
if toolsetsFlag != "" && toolsetsFlag != "default" {
    logger.Fatal("Cannot use both --tools and --toolsets flags together")
}
```

The `--toolsets` persistent flag is declared with default `"all"` in `init.go`. `GetString` returns the default when the user has not passed the flag, so the check fires whenever the operator passes only `--tools`.

Workaround: pass `--toolsets=""` explicitly along with `--tools=...`.

## Suggested fix

Use `pflag.Flag.Changed`, which is `true` only when the operator set the flag on the command line. That maps 1:1 to the intent of the check ("did the user pass `--toolsets`?") without coupling to the declared default.

I have a fix prepared in #380.

sebin:
Fixes #379.

## Summary

`getToolsetsFromCmd` guards against passing both `--tools` and `--toolsets` at the same time. It does so by reading the `--toolsets` value with `GetString` and rejecting any value that is non-empty and not `"default"`:

```go
toolsetsFlag, _ := cmd.Flags().GetString("toolsets")
if toolsetsFlag == "" {
    toolsetsFlag, _ = cmd.Root().PersistentFlags().GetString("toolsets")
}
if toolsetsFlag != "" && toolsetsFlag != "default" {
    logger.Fatal("Cannot use both --tools and --toolsets flags together")
}
```

`--toolsets` is declared with default `"all"` in `init.go`. `GetString` returns the default when the operator has not passed the flag, so the check fires every time someone passes only `--tools`:

```
$ ./terraform-mcp-server --tools=search_providers
time="..." level=fatal msg="Cannot use both --tools and --toolsets flags together"
```

## Change

Replace the value-based check with a `pflag.Flag.Changed` check. `Changed` is `true` only when the operator set the flag on the command line, which is precisely the condition the original check was reaching for. The intent (forbid passing both flags) is preserved; the false positive on the declared default goes away.

Diff is +8 / -5 in `main.go` and a new test file.

## Verification

```bash
$ go build -o ./terraform-mcp-server ./cmd/terraform-mcp-server/

# Before
$ ./terraform-mcp-server stdio --tools=search_providers
level=fatal msg="Cannot use both --tools and --toolsets flags together"
# exit 1

# After
$ ./terraform-mcp-server stdio --tools=search_providers,get_provider_details
level=info msg="Enabled individual tools: [search_providers get_provider_details]"
Terraform MCP Server running on stdio
```

`go test ./...` passes (578/578).

## Test plan

* [x] `TestToolsFlagAloneDoesNotConflictWithDefaultToolsets` — passes only `--tools`, asserts the Fatal is not triggered (uses a logrus `ExitFunc` override so a Fatal call would be observable without exiting the test binary).
* [x] `TestToolsAndToolsetsTogetherTriggersFatal` — locks the negative behavior down: passing both flags explicitly must still Fatal.
* [x] `go test ./...` — 578 pass.
* [x] Manual repro: `./terraform-mcp-server stdio --tools=search_providers` now starts cleanly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ IMPROVEMENTS
 * Run as a non-root user for Kubernetes compatibility. [356] https://github.com/hashicorp/terraform-mcp-server/pull/356
 * Bump go version to 1.26.3 [366] https://github.com/hashicorp/terraform-mcp-server/pull/366
 
+FIXES
+
+* Fix `--tools` alone being falsely flagged as conflicting with `--toolsets` [380](https://github.com/hashicorp/terraform-mcp-server/pull/380)
+
 # 0.5.2
 
 IMPROVEMENTS
--- a/cmd/terraform-mcp-server/main.go
+++ b/cmd/terraform-mcp-server/main.go
@@ -230,12 +230,15 @@ func getToolsetsFromCmd(cmd *cobra.Command, logger *log.Logger) []string {
 	}
 
 	if err == nil && toolsFlag != "" {
-		// Ensure --toolsets is not also set
-		toolsetsFlag, _ := cmd.Flags().GetString("toolsets")
-		if toolsetsFlag == "" {
-			toolsetsFlag, _ = cmd.Root().PersistentFlags().GetString("toolsets")
+		// Ensure --toolsets is not also explicitly set. The flag is declared with
+		// a non-empty default ("all"), so a value-based check would always fire
+		// when only --tools is passed; pflag's Changed field is the precise
+		// "was this flag set on the command line?" signal.
+		toolsetsFlagDef := cmd.Flags().Lookup("toolsets")
+		if toolsetsFlagDef == nil {
+			toolsetsFlagDef = cmd.Root().PersistentFlags().Lookup("toolsets")
 		}
-		if toolsetsFlag != "" && toolsetsFlag != "default" {
+		if toolsetsFlagDef != nil && toolsetsFlagDef.Changed {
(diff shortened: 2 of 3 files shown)


jrhouston (approved):

jrhouston:
Thanks for fixing this @sebin

PR #386, opened 2026-06-22
feat: add organization allowlist gate

leefowlercu:
## summary
- add an HTTP-only organization allowlist gate configured by `--organization-allowlist` or `MCP_ORGANIZATION_ALLOWLIST`
- require `Authorization: Bearer <token>` when the comma-separated allowlist is configured
- validate bearer-token organization access through paginated TFE organization listing without storing the validation client in session state
- document the new comma-separated allowlist configuration and add table-driven coverage

## validation
- `go test ./pkg/client ./cmd/terraform-mcp-server`
- `go test ./...`

### DIFF
--- a/pkg/client/middleware.go
+++ b/pkg/client/middleware.go
@@ -191,13 +191,13 @@ func OrganizationAllowlistMiddleware(allowlist []string, logger *log.Logger) fun
 					http.Error(w, "Terraform token is unauthorized", http.StatusUnauthorized)
 					return
 				}
-				logger.WithError(err).Error("Failed to list Terraform organizations for allowlist validation")
-				http.Error(w, "Failed to validate organization allowlist", http.StatusBadGateway)
+				logger.WithError(err).Error("Failed to validate organization membership for supplied authorization token")
+				http.Error(w, "Failed to validate organization membership for supplied authorization token", http.StatusBadGateway)
 				return
 			}
 			if !allowed {
-				logger.Warn("Rejecting request: Terraform token cannot access an allowlisted organization")
-				http.Error(w, "Terraform token cannot access an allowlisted organization", http.StatusForbidden)
+				logger.Warn("Rejecting request: Supplied authorization token does not have access to any organizations allowed by this server")
+				http.Error(w, "Supplied authorization token does not have access to any organizations allowed by this server", http.StatusForbidden)
 				return
 			}
 
--- a/pkg/client/middleware.go
+++ b/pkg/client/middleware.go
@@ -164,6 +164,12 @@ func OrganizationAllowlistMiddleware(allowlist []string, logger *log.Logger) fun
 				return
 			}
 
+			if r.Header.Get(TerraformAddress) != "" || r.URL.Query().Get(TerraformAddress) != "" {
+				logger.Warn("Rejecting request: Terraform address override is not allowed when organization allowlist is configured")
+				http.Error(w, "Cannot specify Terraform address when organization allowlist is configured", http.StatusForbidden)
+				return
+			}
+
 			token := strings.TrimSpace(getTokenFromAuthHeader(r))
 			if token == "" {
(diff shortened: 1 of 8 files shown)


pkg/client/middleware.go
```diff
@@ -124,6 +143,111 @@ func NewSecurityHandler(handler http.Handler, allowedOrigins []string, corsMode
...
+					http.Error(w, "Terraform token is unauthorized", http.StatusUnauthorized)
+					return
+				}
+				logger.WithError(err).Error("Failed to list Terraform organizations for allowlist validation")
+				http.Error(w, "Failed to validate organization allowlist", http.StatusBadGateway)
+				return
+			}
+			if !allowed {
+				logger.Warn("Rejecting request: Terraform token cannot access an allowlisted organization")
+				http.Error(w, "Terraform token cannot access an allowlisted organization", http.StatusForbidden)
```
jrhouston: Was trying this and found the wording of this error message confusing, maybe we can do:

```
Authorization token does not have access to any organizations allowed by this server 
```

pkg/client/middleware.go
```diff
@@ -124,6 +143,111 @@ func NewSecurityHandler(handler http.Handler, allowedOrigins []string, corsMode
...
+
+			allowed, err := tokenHasAllowedOrganization(r.Context(), lister, allowedOrganizations)
+			if err != nil {
+				if errors.Is(err, tfe.ErrUnauthorized) {
+					logger.Warn("Rejecting request: Terraform token is unauthorized")
+					http.Error(w, "Terraform token is unauthorized", http.StatusUnauthorized)
+					return
+				}
+				logger.WithError(err).Error("Failed to list Terraform organizations for allowlist validation")
+				http.Error(w, "Failed to validate organization allowlist", http.StatusBadGateway)
```
jrhouston: Same with this one, we're not really validating the allowlist – we're using their token to check what organizations they belong to. Perhaps:

```
Failed to validate organization membership for supplied authorization token
```

Above we put the upstream error in the log message so that's useful for debugging.

pkg/client/middleware_test.go
```diff
@@ -140,6 +143,42 @@ func TestLoadCORSConfigFromEnv(t *testing.T) {
...
+			expected: []string{"alpha", "beta", "gamma"},
+		},
+		{
+			name:     "normalizes case",
+			input:    "Alpha,alpha,ALPHA",
+			expected: []string{"alpha", "alpha", "alpha"},
+		},
+		{
+			name:     "blank fields only disable allowlist",
+			input:    " , ,, ",
```
jrhouston: This is a sort of peculiar edge case to support. 

I think they set this flag, the allowlist has a non-zero length, but then it evaluates to empty we should just give them an error? Like `Error: malformed allowlist`.

jrhouston (changes requested):
Tried this out, works good! Left some comments around wording of error messages.

pkg/client/middleware_test.go
leefowlercu: Agreed, better to "fail closed" in this particular scenario

leefowlercu: Updated as of https://github.com/hashicorp/terraform-mcp-server/pull/386/commits/c612fe1e115dd4af9109c38b88d0a14ce2ebdd3d

pkg/client/middleware.go
leefowlercu: Updated in https://github.com/hashicorp/terraform-mcp-server/pull/386/commits/5f27b341455ab75c635efbd7d326673cbbe7ad32

pkg/client/middleware.go
leefowlercu: Updated in https://github.com/hashicorp/terraform-mcp-server/pull/386/commits/5f27b341455ab75c635efbd7d326673cbbe7ad32

jrhouston (approved):

PR #389, opened 2026-07-06
Disable client-supplied TFE_ADDRESS in streamable-http mode

jaylonmcshan19-x:
## PCI review checklist

This PR disables client-supplied `TFE_ADDRESS` in streamable-http mode. Previously a client could set the Terraform address via HTTP header or query parameter, which would redirect the server's requests,  and the `Authorization` bearer token, to an arbitrary endpoint. This removes that capability entirely.

Addresses https://hashicorp.atlassian.net/browse/TFECO-12757 

This PR will introduce a breaking change

The Clients that previously supplied the Terraform address via the `TFE_ADDRESS` header or query parameter (including the "both address and token via headers" pattern) will now receive a `403 err`. The address must now be configured server-side via the `TFE_ADDRESS` env var.

AI assisted with the test cleanup, and updated test messages via find and replace

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 1.0.1
+
+SECURITY
+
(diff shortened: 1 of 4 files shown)


jrhouston (changes requested):
This looks good @jaylonmcshan19-x can you add a CHANGELOG for it

jrhouston (approved):

PR #392, opened 2026-07-07
Add shared secret header

jaylonmcshan19-x:
## PCI review checklist

This PR adds a configurable shared secret that the server sends as the `X-Tf-Mcp-Secret` header on every go-tfe request. 

This allow HCPT/TFE to be able to distinguish requests from a trusted MCP deployment from requests coming from someone else's self-hosted/local instance, so it can enforce IP allowlists on the hosted deployment's traffic.

We already send `X-Forwarded-For`, but that alone doesn't allow HCPT or TFE operators to establish trust. The shared secret provides that identification.

### DIFF
--- a/pkg/client/tfe_client.go
+++ b/pkg/client/tfe_client.go
@@ -24,6 +24,8 @@ const (
 	DefaultTerraformAddress = "https://app.terraform.io"
 	ForwardClientIP         = "MCP_FORWARD_CLIENT_IP"
 	ClientIPKey             = "CLIENT_IP"
+	SharedSecretEnv         = "TF_MCP_SHARED_SECRET"
+	SharedSecretHeader      = "X-Tf-Mcp-Secret"
 )
 
 var activeTfeClients sync.Map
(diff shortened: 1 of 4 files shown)


pkg/client/tfe_client.go
```diff
@@ -72,6 +80,36 @@ func newTfeClient(terraformAddress string, terraformSkipTLSVerify bool, terrafor
...
 }
 
+// the buildTFEConfig func assembles the go-tfe client configuration, including the outbound
+// headers set on every request (User-Agent, optional X-Forwarded-For, and the
+// optional hosted-deployment shared secret).
+//
+// This is split out from newTfeClient so the header logic can be unit-tested (I feel this feature should be tested, so if the header logic gets refactored in the future, we can catch the regression):
+// tfe.NewClient consumes the config and does not expose the headers back, so we can
+// assert against the *tfe.Config this returns instead.
+func buildTFEConfig(terraformAddress string, terraformSkipTLSVerify bool, terraformToken string, clientIP string, logger *log.Logger) *tfe.Config {
```
jrhouston, line 82: Where does this function actually get called? It only seems to be defined here and then called in the test file?

pkg/client/tfe_client.go
```diff
@@ -61,6 +63,12 @@ func newTfeClient(terraformAddress string, terraformSkipTLSVerify bool, terrafor
 	if clientIP != "" {
 		config.Headers.Set("X-Forwarded-For", clientIP)
 	}
+
+	// Attach the shared secret (if configured) so Atlas can identify requests
+	// from a HashiCorp-hosted deployment. Deploy-time static; never logged.
```
jrhouston: This comment is inaccurate. It's not jsut for hashicorp-hosted deployments, it's for TFE too.

jrhouston (changes requested):

jrhouston (approved):

PR #395, opened 2026-07-10
Fix security issues related to session - client mapping, stateless mode session handling and session end cleanup

Shweta-Deshpande:
## PCI review checklist

This PR handles the security vulnerabilities mentioned in:
https://hashicorp.atlassian.net/browse/SECVULN-49459 - Cross-tenant HCP Terraform token reuse in stateless HTTP mode (multi-tenant token bleed)
https://hashicorp.atlassian.net/browse/SECVULN-47947 - Session-scoped TFE client cache bypasses per-token request validation in StreamableHTTP mode

In addition, it also adds a clean way of deleting session limiters for a session when a session unregisters. This functionality replaces the CleanupSessions in ratelimit.go which was never called before.

### DIFF
--- a/pkg/client/tfe_client.go
+++ b/pkg/client/tfe_client.go
@@ -5,6 +5,7 @@ package client
 
 import (
 	"context"
+	"crypto/sha256"
 	"fmt"
 	"net/http"
 	"sync"
@@ -27,15 +28,25 @@ const (
 
 var activeTfeClients sync.Map
 
+type cachedTfeClient struct {
+	client  *tfe.Client
+	token   [32]byte // Store the hash of the token instead of raw value
+	address string
+}
+
 // NewTfeClient creates a new TFE client for the given session
 func NewTfeClient(sessionId string, terraformAddress string, terraformSkipTLSVerify bool, terraformToken string, clientIP string, logger *log.Logger) (*tfe.Client, error) {
 	client, err := newTfeClient(terraformAddress, terraformSkipTLSVerify, terraformToken, clientIP, logger)
 	if err != nil {
 		return nil, err
 	}
-
-	activeTfeClients.Store(sessionId, client)
-	logger.WithField("session_id", sessionId).Info("Created TFE client")
+	// Store the token and address along with the client per session ID
+	activeTfeClients.Store(sessionId, cachedTfeClient{
+		client:  client,
+		token:   sha256.Sum256([]byte(terraformToken)),
+		address: terraformAddress,
+	})
+	logger.Info("Created TFE client")
 	return client, nil
 }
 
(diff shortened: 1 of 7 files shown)


pkg/client/tfe_client.go
```diff
@@ -27,15 +28,25 @@ const (
...
 func NewTfeClient(sessionId string, terraformAddress string, terraformSkipTLSVerify bool, terraformToken string, clientIP string, logger *log.Logger) (*tfe.Client, error) {
 	client, err := newTfeClient(terraformAddress, terraformSkipTLSVerify, terraformToken, clientIP, logger)
 	if err != nil {
 		return nil, err
 	}
-
-	activeTfeClients.Store(sessionId, client)
-	logger.WithField("session_id", sessionId).Info("Created TFE client")
+	// Store the token and address along with the client per session ID
+	activeTfeClients.Store(sessionId, cachedTfeClient{
```
jrhouston, line 43: I think since we are creating a hash of the token we can use that as the key that maps to the client.

pkg/client/tfe_client.go
```diff
@@ -27,15 +28,25 @@ const (
 
 var activeTfeClients sync.Map
 
+type cachedTfeClient struct {
+	client  *tfe.Client
+	token   [32]byte // Store the hash of the token instead of raw value
+	address string
```
jrhouston: Storing the address is now superfluous as we've removed the ability for clients to set the address via header anyway.

pkg/client/tfe_client.go
```diff
@@ -92,12 +103,33 @@ func GetTfeClientFromContext(ctx context.Context, logger *log.Logger) (*tfe.Clie
...
+	// Check if the cached session ID's token+address match the current token+address
+	if value, ok := activeTfeClients.Load(session.SessionID()); ok {
+		cachedClient := value.(cachedTfeClient)
+		currentTokenHash := sha256.Sum256([]byte(currentToken))
+		if cachedClient.token == currentTokenHash && cachedClient.address == currentAddress {
+			return cachedClient.client, nil
+		}
+		// Current request token and address not found in cache. Delete the session ID from the sync map.
+		activeTfeClients.Delete(session.SessionID())
+	}
```
jrhouston, line 130: If we just use the token hash as the cache key instead of sessionID, and we're not storing the address anymore then we wont need this logic.

jrhouston (changes requested):

pkg/client/tfe_client.go
Shweta-Deshpande: I think that might not work since we are using the sessionID as an arg to DeleteTfeClient() and we need the sessionID to find the right entry to delete. Is the token available at EndSessionHandler?
This will force us to have a struct mapping token and sessionID then.

pkg/client/tfe_client.go
Shweta-Deshpande: Done

pkg/client/tfe_client.go
Shweta-Deshpande: We might need to keep the sessionID: tokenHash mapping.

jrhouston (dismissed):

jrhouston (approved):

PR #396, opened 2026-07-11
fix: prevent organization allowlist token bypass

leefowlercu:
## What was wrong

When the organization allowlist was enabled, the server checked the token from the `Authorization: Bearer ...` header to decide whether a request could access an allowed organization.

However, the request context preferred the token from the `TFE_TOKEN` header. A caller could provide an allowed token in the Authorization header and a different token in the `TFE_TOKEN` header. The request would pass the allowlist check, but Terraform API operations would use the different, unchecked token.

## What changed

The Authorization bearer token now takes precedence over the `TFE_TOKEN` header. This makes the token checked by the organization allowlist the same token used for downstream Terraform API requests.

Bearer tokens are also trimmed of surrounding whitespace so the value checked by the allowlist exactly matches the value stored in the request context. Requests can still use the `TFE_TOKEN` header as a fallback when no bearer token is present and the organization allowlist is not enabled.

The README now documents the precedence when both headers are provided.

## How this is tested

`TestTerraformContextMiddleware` provides different values in the Authorization and `TFE_TOKEN` headers and verifies that the request context receives the bearer token. This proves downstream Terraform clients will select the token that the allowlist checks.

`TestOrganizationAllowlistUsesValidatedTokenDownstream` composes the Terraform context and organization allowlist middleware in the same order used by the HTTP server. It sends an allowed bearer token together with a different `TFE_TOKEN`, then verifies two things:

- The fake Terraform API receives the allowed bearer token when the server checks organization membership.
- The wrapped downstream handler receives that same allowed token instead of the conflicting `TFE_TOKEN` value.

The bearer-token parsing tests also verify that blank tokens are rejected and surrounding whitespace is removed consistently.

The complete test suite passes with:

```text
go test ./... -count=1
```

### DIFF
--- a/pkg/client/middleware_test.go
+++ b/pkg/client/middleware_test.go
@@ -495,6 +495,52 @@ func TestOrganizationAllowlistMiddleware(t *testing.T) {
 	}
 }
 
+func TestOrganizationAllowlistUsesValidatedTokenDownstream(t *testing.T) {
+	logger := log.New()
+	logger.SetLevel(log.ErrorLevel)
+
+	upstreamAuthorization := make(chan string, 1)
+	terraformServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.URL.Path == "/api/v2/organizations" {
+			select {
+			case upstreamAuthorization <- r.Header.Get("Authorization"):
+			default:
+			}
+		}
+		w.Header().Set("Content-Type", "application/vnd.api+json")
+		_, err := io.WriteString(w, `{"data":[{"id":"alpha","type":"organizations","attributes":{"name":"alpha"}}]}`)
+		assert.NoError(t, err)
+	}))
+	t.Cleanup(terraformServer.Close)
+	t.Setenv(TerraformAddress, terraformServer.URL)
+
+	var downstreamToken string
+	next := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		downstreamToken, _ = r.Context().Value(contextKey(TerraformToken)).(string)
+		w.WriteHeader(http.StatusOK)
+	})
+
+	handler := TerraformContextMiddleware(logger)(
+		OrganizationAllowlistMiddleware([]string{"alpha"}, logger)(next),
+	)
+
+	req := httptest.NewRequest(http.MethodPost, "/mcp", nil)
+	req.Header.Set("Authorization", "Bearer allowed-token")
+	req.Header.Set(TerraformToken, "different-token")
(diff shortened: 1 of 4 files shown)


pkg/client/middleware_test.go
```diff
@@ -495,6 +495,52 @@ func TestOrganizationAllowlistMiddleware(t *testing.T) {
...
+	logger := log.New()
+	logger.SetLevel(log.ErrorLevel)
+
+	upstreamAuthorization := make(chan string, 1)
+	terraformServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.URL.Path == "/api/v2/organizations" {
+			select {
+			case upstreamAuthorization <- r.Header.Get("Authorization"):
+			default:
+			}
```
jrhouston: The use of channels/select is kinda unnecessary here, what about this:

```go
func TestOrganizationAllowlistUsesValidatedTokenDownstream(t *testing.T) {
	logger := log.New()
	logger.SetLevel(log.ErrorLevel)

	var upstreamToken string
	terraformServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		if r.URL.Path == "/api/v2/organizations" {
			upstreamToken = r.Header.Get("Authorization")
		}
		w.Header().Set("Content-Type", "application/vnd.api+json")
		_, err := io.WriteString(w, `{"data":[{"id":"alpha","type":"organizations","attributes":{"name":"alpha"}}]}`)
		assert.NoError(t, err)
	}))
	t.Cleanup(terraformServer.Close)
	t.Setenv(TerraformAddress, terraformServer.URL)

	var downstreamToken string
	next := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		downstreamToken, _ = r.Context().Value(contextKey(TerraformToken)).(string)
		w.WriteHeader(http.StatusOK)
	})

	handler := TerraformContextMiddleware(logger)(
		OrganizationAllowlistMiddleware([]string{"alpha"}, logger)(next),
	)

	req := httptest.NewRequest(http.MethodPost, "/mcp", nil)
	req.Header.Set("Authorization", "Bearer allowed-token")
	req.Header.Set(TerraformToken, "disallowed-token")
	rr := httptest.NewRecorder()

	handler.ServeHTTP(rr, req)

	assert.Equal(t, http.StatusOK, rr.Code)
	assert.Equal(t, "Bearer allowed-token", upstreamToken)
	assert.Equal(t, "allowed-token", downstreamToken)
}
```

jrhouston (changes requested):

jrhouston (approved):

PR #415, opened 2026-07-22
fix: bump x/text, x/net, grpc to resolve security vunlnerabilities

glbyktjys:
Bumps three Go dependencies to address OSV-reported vulnerabilities found during the release binary scan:

- golang.org/x/text from v0.37.0 to v0.40.0
- golang.org/x/net from v0.55.0 to v0.57.0
- google.golang.org/grpc from v1.81.1 to v1.82.1

golang.org/x/sync and golang.org/x/sys were also updated by `go mod tidy`.

Verified clean with `govulncheck ./....`

No changelog needed.


jrhouston (approved):
Thanks for updating these @glbyktjys!

PR #420, opened 2026-07-24
Add `create_project` and `delete_project`

glbyktjys:
## Description

Adds `create_project` and `delete_project` tools to the Terraform toolset for managing TFC/TFE projects.

## Changes

- [New Tool] `create_project` Creates a new TFC/TFE project in a specified organization. Requires `terraform_org_name` and `project_name`; supports an optional description.
- [New Tool] `delete_project` Deletes a TFC/TFE project by ID. Requires `project_id`.
- Both tool gated behind the `ENABLE_TF_OPERATIONS`

## Testing
Added unit tests for `create_project` and `delete_project`.
Tested locally and manually verified project creation and deletion against TFC using an AI agent (IBM Bob).


pkg/tools/tfe/create_project.go
```diff
@@ -0,0 +1,95 @@
...
+		Name: projectName,
+	}
+
+	if description != "" {
+		options.Description = &description
+	}
+
+	project, err := tfeClient.Projects.Create(ctx, terraformOrgName, options)
+	if err != nil {
+		return ToolErrorf(logger, "failed to create project '%s' in org '%s': %v", projectName, terraformOrgName, err)
```
jrhouston: You can use `%q` as the formatting directive here instead of quoting things by hand. 

See Go's formatting directives here: https://pkg.go.dev/fmt#hdr-Printing

pkg/tools/tfe/create_project.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	projectJSON, err := json.Marshal(&ProjectSummary{
+		ID:   project.ID,
+		Name: project.Name,
+	})
```
jrhouston: Because [`ProjectSummary`](https://github.com/hashicorp/terraform-mcp-server/blob/main/pkg/tools/tfe/list_terraform_projects.go#L90-L93) only has these two attributes anyway, you could just do:

```go
&ProjectSumary{project.Id, project.Name}
```

pkg/tools/tfe/create_project.go
```diff
@@ -0,0 +1,95 @@
...
+				mcp.Required(),
+				mcp.Description("The project name. Must be 3-40 characters, contain only letters, numbers, spaces, hyphens, and underscores, and not start or end with a space."),
+				mcp.MinLength(3),
+				mcp.MaxLength(40),
+				mcp.Pattern(`^[A-Za-z0-9_-][A-Za-z0-9_-]*[A-Za-z0-9_-]$`),
+			),
+			mcp.WithString("description",
+				mcp.Description("Optional project description. Must be no more than 256 characters"),
+				mcp.MaxLength(256),
+			),
```
jrhouston, line 46: At project creation time you can also specify the default execution mode for workspaces in the project, e.g. `remote`, `local`, or `agent` so it probably sense to add a `default_execution_mode` parameter here so we can do prompts like:

> Create a new project called `jioayue-demo-project` which defaults to local execution.

It also lets you specify a default agent pool, but we don't have any tools for managing agent pools yet so we can come back and add that later.

pkg/tools/tfe/delete_project.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/mark3labs/mcp-go/mcp"
+	"github.com/mark3labs/mcp-go/server"
+)
+
+// DeleteProject creates a tool to delete a Terraform project by ID.
+// TFC/TFE will reject the request if the project still contains workspaces or stacks.
+func DeleteProject(logger *log.Logger) server.ServerTool {
+	return server.ServerTool{
+		Tool: mcp.NewTool("delete_project",
+			mcp.WithDescription(`Deletes a Terraform project by ID. This is a destructive operation. The request will fail if the project still contains workspaces or stacks.`),
```
jrhouston: You can use this description to tell the model to call the `list_terraform_projects` tool so the user doesn't have to figure out the project ID by hand.

pkg/tools/tfe/delete_project.go
```diff
@@ -0,0 +1,57 @@
...
+	projectID = strings.TrimSpace(projectID)
+
+	tfeClient, err := client.GetTfeClientFromContext(ctx, logger)
+	if err != nil {
+		return ToolError(logger, "failed to get Terraform client - ensure TFE_TOKEN and TFE_ADDRESS are configured", err)
+	}
+
+	err = tfeClient.Projects.Delete(ctx, projectID)
+	if err != nil {
+		return ToolErrorf(logger, "failed to delete project '%s': %v", projectID, err)
```
jrhouston: Use `%q` here too

pkg/tools/tfe/delete_project.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+
+	return mcp.NewToolResultText(fmt.Sprintf("project %s deleted", projectID)), nil
```
jrhouston: Use `%q` here too

pkg/tools/tfe/create_project.go
```diff
@@ -0,0 +1,95 @@
...
+	return server.ServerTool{
+		Tool: mcp.NewTool("create_project",
+			mcp.WithDescription(`Creates a new Terraform project in the specified organization. Call list_terraform_orgs first, then use ask_followup_question with each org name as a suggestion button so the user can select one without typing. Then ask the user for the project name and if they'd like to add an optional description — wait for their answer before proceeding.`),
+			mcp.WithTitleAnnotation("Create a new Terraform project"),
+			mcp.WithOpenWorldHintAnnotation(true),
+			mcp.WithReadOnlyHintAnnotation(false),
+			mcp.WithDestructiveHintAnnotation(false),
+			mcp.WithString("terraform_org_name",
+				mcp.Required(),
+				mcp.Description("The name of the Terraform Cloud/Enterprise organization to create the project in"),
```
jrhouston: Side note: we should probably make a constant with this description because this parameter appears in a ton of tools, and it currently is not consistent across the tool definitions.

pkg/tools/tfe/create_project.go
```diff
@@ -0,0 +1,95 @@
...
+
+	"github.com/mark3labs/mcp-go/mcp"
+	"github.com/mark3labs/mcp-go/server"
+)
+
+// CreateProject creates a tool to create a new Terraform project.
+func CreateProject(logger *log.Logger) server.ServerTool {
+	return server.ServerTool{
+		Tool: mcp.NewTool("create_project",
+			mcp.WithDescription(`Creates a new Terraform project in the specified organization. Call list_terraform_orgs first, then use ask_followup_question with each org name as a suggestion button so the user can select one without typing. Then ask the user for the project name and if they'd like to add an optional description — wait for their answer before proceeding.`),
```
jrhouston: `ask_follow_up_question` isn't actually a tool in this project, and elicitation is currently one of the most inconsistently implemented features of the MCP spec across clients. I'd worthsmith this description a little more conservatively for now.

jrhouston (changes requested):

pkg/tools/tfe/create_project.go
glbyktjys: I created a `description.go` file and added a `terraformOrgNameDescription` constant. Would it be okay to centralize reusable input descriptions this way, so future descriptions for fields like `project_id`, `workspace_id`, and `workspace_name` can also be placed there?

pkg/tools/tfe/create_project.go
glbyktjys: I've removed the `ask_followup_question` reference and softened the description a bit to just point at `list_terraform_orgs`.

The original wording trying to make an attempt of ask optional description parameter before creating. Is one-shot creation(like using required fields only and silently omit anything that user didn't mention) is the behavior we preferred here?

pkg/tools/tfe/create_project.go
glbyktjys: Sounds good. I’ve added `default_execution_mode` as an optional parameter.

I'll follow up with agent pool management tools and the remaining ProjectCreateOptions fields (e.g. default_agent_pool_id) in a separate PR.

pkg/tools/tfe/create_project.go
```diff
@@ -0,0 +1,110 @@
...
+		case "local":
+			options.DefaultExecutionMode = tfe.String("local")
+		case "agent":
+			options.DefaultExecutionMode = tfe.String("agent")
+		case "remote":
+			options.DefaultExecutionMode = tfe.String("remote")
+		default:
+			return ToolErrorf(logger, "invalid default_execution_mode %q - must be 'local', 'agent', 'remote'", defaultExecutionMode)
+		}
+	}
```
jrhouston, line 96: I'm not sure we need to write out each of these cases explicitly.

Instead of a switch,  I would do something like:

```golang
if !slice.Contains([]{"local", "agent", "remote"}, defaultExecutionMode) {
   // return the error
}
options.DefaultExecutionMode = tfe.String(defaultExecutionMode)
```

You could also use a map like this, but it's kind of a micro-optimization:

```golang
validModes := map[string]struct{}{
    "local": {},
    "remote": {},
    "agent": {}
}

if _, valid := validModes[defaultExecutionMode]; !valid {
    // return the error
}
options.DefaultExecutionMode = tfe.String(defaultExecutionMode)
```

pkg/tools/tfe/create_project_test.go
```diff
@@ -0,0 +1,226 @@
...
+		var decoded ProjectSummary
+		require.NoError(t, json.Unmarshal(data, &decoded))
+		assert.Equal(t, "prj-abc123", decoded.ID)
+		assert.Equal(t, "my-project", decoded.Name)
+
+		// Verify JSON key names match the spec
+		assert.Contains(t, string(data), `"project_id"`)
+		assert.Contains(t, string(data), `"project_name"`)
+	})
+}
```
jrhouston: These tests don't actually test the application code  – just remove them.

jrhouston (changes requested):
Couple things

pkg/tools/tfe/create_project.go
```diff
@@ -0,0 +1,110 @@
...
+	return server.ServerTool{
+		Tool: mcp.NewTool("create_project",
+			mcp.WithDescription(`Creates a new Terraform project in the specified organization.`),
+			mcp.WithTitleAnnotation("Create a new Terraform project"),
+			mcp.WithOpenWorldHintAnnotation(true),
+			mcp.WithReadOnlyHintAnnotation(false),
+			mcp.WithDestructiveHintAnnotation(false),
+			mcp.WithString("terraform_org_name",
+				mcp.Required(),
+				mcp.Description(terraformOrgNameDescription+"to create the project in"),
```
jrhouston:
```suggestion
				mcp.Description(terraformOrgNameDescription+" to create the project in"),
```

pkg/tools/tfe/descriptions.go
```diff
@@ -0,0 +1,6 @@
+// Copyright IBM Corp. 2025
```
jrhouston, line 1: maybe call this file `descriptions.go` instead?

If you want to be a good citizen you can go and update all the tool definitions that have `terraform_org_name` and replace the hardcoded text with this const. 😄

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ FEATURES
 
 * [New Tool] `list_state_versions` Lists all state versions for a given workspace. Requires `terraform_org_name` and `workspace_name`; supports optional pagination params.
 * [New Tool] `get_state_version` Retrieves a Terraform state version. If `state_version_id` is provided, retrieves that specific state version. Otherwise, retrieves the latest state version for the specified `workspace_id`. One of `state_version_id` or `workspace_id` must be provided.
+* [New Tool] `create_project` Creates a new Terraform project in the specified organization. Requires `terraform_org_name` and `project_name`.
```
jrhouston: While you're here can you change the heading to `# 1.2.0` since we're adding a bunch of tools.

jrhouston (changes requested):
couple more little things.

glbyktjys:
Hi @jrhouston @jaylonmcshan19-x, I’ve addressed the feedback, and the PR is ready for another review. Thank you!

pkg/tools/tfe/create_project.go
```diff
@@ -0,0 +1,104 @@
...
+		Handler: func(ctx context.Context, request mcp.CallToolRequest) (*mcp.CallToolResult, error) {
+			return createProjectHandler(ctx, request, logger)
+		},
+	}
+}
+
+func createProjectHandler(ctx context.Context, request mcp.CallToolRequest, logger *log.Logger) (*mcp.CallToolResult, error) {
+	terraformOrgName, err := request.RequireString("terraform_org_name")
+	if err != nil {
+		return ToolError(logger, "missing required input: terraform_org_name", err)
```
jrhouston, line 60: SIde note, not for this PR: we could totally DRY this up and just have a function like `MissingInputError("terraform_org_name")`

pkg/tools/tfe/create_project.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	if description != "" {
+		options.Description = &description
+	}
+
+	if defaultExecutionMode != "" {
+		if !slices.Contains([]string{"local", "agent", "remote"}, strings.ToLower(defaultExecutionMode)) {
+			return ToolErrorf(logger, "invalid default_execution_mode %q - must be 'local', 'agent', 'remote'", defaultExecutionMode)
+		}
```
jrhouston, line 94: You hard code this list twice, once in the slice and once in the error message. You can make the slice a variable, then use `strings.Join()` in the error message so you don't have to repeat yourself.

edit: actually 3 times because it's in the parameter description too! maybe make it a const at the top of the file?

pkg/tools/tfe/create_project.go
glbyktjys: That makes sense. I can follow up with a refactor PR to standardize missing-input errors and similar repetitive error handling patterns across the tools.

jrhouston (changes requested):
One tiny thing and then we're good

jrhouston (approved):

PR #422, opened 2026-07-27
Add `delete_team` tool

Marin2409:
## Changes Overview:

New Tool added → `delete_team`: Permanently deletes a Terraform team by its `team_id`.

## What changed
`pkg/tools/tfe/delete_team.go` — tool definition and handler
`pkg/tools/tfe/delete_team_test.go` — unit tests
`pkg/tools/dynamic_tool.go` — registers delete_team behind ENABLE_TF_OPERATIONS
`pkg/toolsets/mapping.go` — adds delete_team to the terraform toolset
`pkg/tools/tfe/action_run.go` — replaced JSON map response with plain text confirmation, removed unused encoding/json import
`pkg/tools/tfe/force_unlock_workspace.go` — replaced JSON map response with plain text confirmation, removed unused encoding/json import
`CHANGELOG.md` — entry added

## Inputs / Output
Required: `team_id` (e.g. team-abc123def456)
Returns:
```go
"Team %q deleted", teamID
```

## PCI review checklist

### DIFF
--- a/pkg/tools/tfe/delete_team.go
+++ b/pkg/tools/tfe/delete_team.go
@@ -5,6 +5,8 @@ package tools
 
 import (
 	"context"
+	"encoding/json"
+	"fmt"
 	"strings"
 
 	"github.com/hashicorp/terraform-mcp-server/pkg/client"
@@ -20,8 +22,8 @@ func DeleteTeam(logger *log.Logger) server.ServerTool {
 	return server.ServerTool{
 		Tool: mcp.NewTool(
 			"delete_team",
-			mcp.WithDescription(`Permanently deletes a Terraform team by its team_id.  If you don't have the team_id, look it up first via the organization's Teams URL page from the HCP Terraform/TFE UI.`),
-			mcp.WithTitleAnnotation("Deletes a Terraform Team by team_id"),
+			mcp.WithDescription("Permanently deletes a Terraform team by its team_id.  If you don't have the team_id, look it up first via the organization's Teams URL page from the HCP Terraform/TFE UI."),
+			mcp.WithTitleAnnotation(`Deletes a Terraform Team by team_id`),
 			mcp.WithReadOnlyHintAnnotation(false),
 			mcp.WithOpenWorldHintAnnotation(true),
 			mcp.WithDestructiveHintAnnotation(true),
@@ -34,13 +36,11 @@ func DeleteTeam(logger *log.Logger) server.ServerTool {
 			return deleteTeamHandler(ctx, request, logger)
 		},
 	}
-
 }
 
 // deleteTeamHandler handles tool logics and functionality
 func deleteTeamHandler(ctx context.Context, request mcp.CallToolRequest, logger *log.Logger) (*mcp.CallToolResult, error) {
 
-	// Init clint object
 	tfeClient, err := client.GetTfeClientFromContext(ctx, logger)
 	if err != nil {
 		return ToolError(logger, "Failed to get Terraform client", err)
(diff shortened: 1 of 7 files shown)


pkg/tools/tfe/delete_team.go
```diff
@@ -0,0 +1,78 @@
...
+		return ToolError(logger, "Failed to get Terraform client", err)
+	}
+	if tfeClient == nil {
+		return ToolError(logger, "Failed to get Terraform client - ensure TFE_TOKEN and TFE_ADDRESS are configured", nil)
+	}
+
+	team, err := tfeClient.Teams.Read(ctx, teamID)
+	if err != nil {
+		return ToolErrorf(logger, "Team not found: %s", teamID)
+	}
```
jrhouston: Not sure this extra call to check the team exists is necessary. I assume the call below will fail if the team doesn't exist?

pkg/tools/tfe/delete_team.go
```diff
@@ -0,0 +1,78 @@
...
+
+	err = tfeClient.Teams.Delete(ctx, teamID)
+	if err != nil {
+		return ToolErrorf(logger, "Failed to delete team %q", teamID)
+	}
+
+	result := map[string]interface{}{
+		"success": true,
```
jrhouston: I think we need to come up with a standard format for these "success message" tool results because they're not consistent across tools. 

Returning `"success": true` is probably just wasting tokens because the tool call result already contains whether or not it succeeded or error'd

In this case returning the `team_id` is also not needed because the model already knows the team_id because it used it to call the tool.

pkg/tools/tfe/delete_team_test.go
```diff
@@ -0,0 +1,122 @@
...
+
+		for _, tt := range tests {
+			t.Run(tt.name, func(t *testing.T) {
+				cleaned := strings.TrimLeft(strings.TrimSpace(tt.input), "/")
+				assert.Equal(t, tt.expected, cleaned)
+			})
+		}
+	})
+
+}
```
jrhouston: tests don't actually test the application code

pkg/tools/tfe/delete_team.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+// deleteTeamHandler handles tool logics and functionality
+func deleteTeamHandler(ctx context.Context, request mcp.CallToolRequest, logger *log.Logger) (*mcp.CallToolResult, error) {
+
+	teamID, err := request.RequireString("team_id")
+	if err != nil {
+		return ToolError(logger, "Missing required input: team_id", err)
+	}
+	teamID = strings.TrimLeft(strings.TrimSpace(teamID), "/")
```
jrhouston: Why the need to trim `/` here?

jrhouston (changes requested):

pkg/tools/tfe/delete_team_test.go
Marin2409: test cases removed

pkg/tools/tfe/delete_team.go
Marin2409: I initially added the TrimLeft(..., "/") because, as far as I know, the easiest way to get a team_id manually from the Terraform Cloud UI is by copying it from the URL. I thought that maybe an user might accidentally include the leading "/" when copying it. This is only the case where the `team_id` isn't already in the model's context and the user is retrieving it manually.

Screenshot 2026-08-03 at 11 58 48 AM

Ill still remove it, let me know if any feedback.

pkg/tools/tfe/delete_team.go
Marin2409: Good catch. Changed it to 1 call only.

pkg/tools/tfe/delete_team.go
Marin2409: Make sense. Updated all three action tools to drop the JSON wrapper and return plain text confirmations, following the `delete_project.go` pattern. Removed "success": true (which is redundant with NewToolResultText) and stopped echoing back IDs the model already has.

PR #424, opened 2026-07-27
Add `add_team_member` tool

Marin2409:
## Changes Overview

New Tool → `add_team_member`: Adds a single member to a Terraform Cloud/Enterprise team by either username or organization membership ID.

## What changed
- `pkg/tools/tfe/add_team_member.go` — Tool definition and handler
- `pkg/tools/tfe/add_team_member_test.go` — Unit tests
- `pkg/tools/dynamic_tool.go` — Registers `add_team_member`
- `pkg/toolsets/mapping.go` — Adds `add_team_member` to the Terraform toolset
- `CHANGELOG.md` — Entry added

## Inputs
- Required: `team_id` (e.g. `team-abc123def456`)
- Optional: `username` — accepted-invite users only
- Optional: `organization_membership_id` — works for both pending and accepted invites

Exactly one of `username` or `organization_membership_id` must be provided. Providing both is a validation error.

## Output

**Success:**
```
"Successfully added member to team \"team-abc123def456\""
```

**Failure:** tool call errors with a descriptive message, e.g.:
```
"Failed to add member \"alice\" to team \"team-abc123def456\": ..."
```

## PCI review checklist

### DIFF
--- a/pkg/tools/tfe/add_team_member.go
+++ b/pkg/tools/tfe/add_team_member.go
@@ -1,0 +1,46 @@
+// Copyright IBM Corp. 2025
+// SPDX-License-Identifier: MPL-2.0
+
+package tools
+
+import (
+	"context"
+
+	"github.com/mark3labs/mcp-go/mcp"
+	"github.com/mark3labs/mcp-go/server"
+	log "github.com/sirupsen/logrus"
+)
+
+// AddTeamMemeber creates a tool to add a new member to your Terraform team.
+func AddTeamMemeber(logger *log.Logger) server.ServerTool {
+	return server.ServerTool{
+		Tool: mcp.NewTool(
+			"add_team_member",
+			mcp.WithDescription("Adds member’s to a team. This is a write operation"),
+			mcp.WithTitleAnnotation(`Adds member’s to a team`),
+			mcp.WithOpenWorldHintAnnotation(true),
+			mcp.WithReadOnlyHintAnnotation(false),
+			mcp.WithDestructiveHintAnnotation(false),
+			mcp.WithString("team_id",
+				mcp.Description("The Teams's id (e.g., 'team-abc123def456')"),
+			),
+			mcp.WithString("username",
+				mcp.Description("The Member's username"),
+			),
+			mcp.WithString("organization_membership_ids",
+				mcp.Description("The Org. membership's id"),
+			),
+		),
+
+		Handler: func(ctx context.Context, request mcp.CallToolRequest) (*mcp.CallToolResult, error) {
+			return addTeamMemberHandler(ctx, request, logger)
+
+		},
+	}
+}
+
+// addTeamMemberHandler handles tool logics and functionality
+func addTeamMemberHandler(ctx context.Context, request mcp.CallToolRequest, logger *log.Logger) (*mcp.CallToolResult, error) {
+
+	return mcp.NewToolResultText(""), nil
+}
--- a/pkg/tools/tfe/add_team_member.go
+++ b/pkg/tools/tfe/add_team_member.go
@@ -5,7 +5,12 @@ package tools
 
 import (
 	"context"
+	"encoding/json"
+	"fmt"
+	"strings"
 
+	"github.com/hashicorp/go-tfe"
+	"github.com/hashicorp/terraform-mcp-server/pkg/client"
 	"github.com/mark3labs/mcp-go/mcp"
 	"github.com/mark3labs/mcp-go/server"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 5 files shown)


pkg/tools/tfe/add_team_member.go
```diff
@@ -0,0 +1,126 @@
...
+			mcp.WithDescription("Adds one or more members to a Terraform Cloud/Enterprise team. Members can be identified by username (accepted invites only) or by organization membership ID (accepted and pending invites). Both can be provided in a single call."),
+			mcp.WithTitleAnnotation(`Add members to a Terraform team`),
+			mcp.WithOpenWorldHintAnnotation(true),
+			mcp.WithReadOnlyHintAnnotation(false),
+			mcp.WithDestructiveHintAnnotation(false),
+			mcp.WithString("team_id",
+				mcp.Required(),
+				mcp.Description("The ID of the Terraform Cloud/Enterprise team to add members to (e.g., 'team-abc123def456')"),
+			),
+			mcp.WithString("username",
```
jrhouston, line 32: The tool is called add `add_team_member` but it allows you to add more than one. Either we should call it `add_team_members` or make it only accept one user at a time.

pkg/tools/tfe/add_team_member.go
```diff
@@ -0,0 +1,126 @@
...
+func addTeamMemberHandler(ctx context.Context, request mcp.CallToolRequest, logger *log.Logger) (*mcp.CallToolResult, error) {
+
+	teamID, err := request.RequireString("team_id")
+	if err != nil {
+		return ToolError(logger, "Missing required input: team_id", err)
+	}
+	username := request.GetString("username", "")
+	organizationMembershipID := request.GetString("organization_membership_ids", "")
+
+	teamID = strings.TrimLeft(strings.TrimSpace(teamID), "/")
```
jrhouston: Why do we trim `/` here? the format is `team-xxxx`

pkg/tools/tfe/add_team_member.go
```diff
@@ -0,0 +1,126 @@
...
+	username = strings.TrimSpace((username))
+	organizationMembershipID = strings.TrimSpace(organizationMembershipID)
+
+	var usernames []string
+	if username != "" {
+		usernames = strings.Split(username, ",")
+		for i, u := range usernames {
+			usernames[i] = strings.TrimSpace(u)
+		}
+	}
```
jrhouston: ~Why do we need to built this slice when the API call just takes a CSV string anyway?~

edit: nvm i read the code incorrectly

pkg/tools/tfe/add_team_member.go
```diff
@@ -0,0 +1,126 @@
...
+	return mcp.NewToolResultText(string(resultJSON)), nil
+}
+
+// AddMemberSummary is a truncated summary of Added Members details for listing
+type AddMemberSummary struct {
+	Success             bool     `json:"success"`
+	AddedByUsername     []string `json:"added_by_username,omitempty"`
+	AddedByMembershipID []string `json:"added_by_membership_id,omitempty"`
+	Errors              []string `json:"errors,omitempty"`
```
jrhouston: I'm not sure I like this duality where the tool can partially succeed and partially fail? Like some users get added and some might not and the tool call still succeeds but the ones who didn't get added are captured in the output.

I'm inclined to keep it simple for now and have the tool just add one user at a time, and the tool call errors if they aren't added.

pkg/tools/tfe/add_team_member_test.go
```diff
@@ -0,0 +1,174 @@
...
+				if tt.expected == nil {
+					assert.Nil(t, result)
+				} else {
+					require.Len(t, result, len(tt.expected))
+					assert.Equal(t, tt.expected, result)
+				}
+			})
+		}
+	})
+}
```
jrhouston: doesn't actually test the application code

jrhouston (changes requested):

pkg/tools/tfe/add_team_member.go
Marin2409: I initially added the TrimLeft(..., "/") because, as far as I know, the easiest way to get a team_id manually from the Terraform Cloud UI is by copying it from the URL. I thought that maybe an user might accidentally include the leading "/" when copying it. This is only the case where the `team_id` isn't already in the model's context and the user is retrieving it manually.

Screenshot 2026-08-03 at 11 58 48 AM

Ill still remove it, let me know if any feedback.

pkg/tools/tfe/add_team_member_test.go
Marin2409: Test case removed

pkg/tools/tfe/add_team_member.go
Marin2409: Kept as `add_team_member`. Tool logic change to one user at a time.

pkg/tools/tfe/add_team_member.go
Marin2409: Good call, updated. The tool now accepts a single username or organization_membership_id per call. No partial-success state anymore. Now, for multiple members the model just calls it once per person.

PR #427, opened 2026-07-28
add `create_team`

glbyktjys:
Adds a `create_team` tool to the Terraform toolset for creating teams in Terraform Cloud or Terraform Enterprise organizations.

## Changes
- `create_team.go`: `CreateTeam` and `createTeamHandler`
- `create_team_test.go`: unit tests
- `mapping.go` and `dynamic_tool`: register create_team with Terraform toolset and lazily adds it to the MCP server when enabled and available for an authenticated TFE session.
- CHANGELOG.md

## Testing

- Added unit tests for tool creation and parameters validation.
- Tested locally through the MCP AI Agent. The tool reaches the TFE API correctly but creation fails only because team management requires HCP Terraform Plus tier, not due to any implementation issue.

### DIFF
--- a/.github/workflows/hcpt_test.yml
+++ b/.github/workflows/hcpt_test.yml
@@ -35,5 +35,6 @@ jobs:
         env:
           TF_MCP_ENDPOINT: http://localhost:8080/mcp
           TFE_TOKEN: ${{ secrets.TFE_TOKEN_TESTING }}
+          TFE_ORG_NAME: ${{ secrets.TEST_TF_ORG_NAME }}
         run: |
           make test-hcpt
--- a/.github/workflows/hcpt_test.yml
+++ b/.github/workflows/hcpt_test.yml
@@ -35,6 +35,6 @@ jobs:
         env:
           TF_MCP_ENDPOINT: http://localhost:8080/mcp
           TFE_TOKEN: ${{ secrets.TFE_TOKEN_TESTING }}
-          TFE_ORG_NAME: ${{ secrets.TEST_TF_ORG_NAME }}
+          TFE_ORG_NAME: ${{ vars.TEST_TF_ORG_NAME }}
         run: |
           make test-hcpt
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ FEATURES
 * [New Tool] `get_run_comments` Lists all discussion comments associated with a given Terraform run. Requires `run_id`.
 * [New Tool] `create_project` Creates a new Terraform project in the specified organization. Requires `terraform_org_name` and `project_name`. [420](https://github.com/hashicorp/terraform-mcp-server/pull/420)
 * [New Tool] `delete_project` Deletes a Terraform project by ID. Requires `project_id`. TFC/TFE will refuse to delete a project that still contains workspaces or stacks. [420](https://github.com/hashicorp/terraform-mcp-server/pull/420)
+* [New Tool] `create_team` Creates a new team in a Terraform Cloud/Enterprise organization. Requires `terraform_org_name` and `team_name`; optional `visibility` ("secret" or "organization"). [427](https://github.com/hashicorp/terraform-mcp-server/pull/427)
 
 IMPROVEMENTS
 
--- a/pkg/tools/dynamic_tool.go
+++ b/pkg/tools/dynamic_tool.go
@@ -116,6 +116,22 @@ func (r *DynamicToolRegistry) registerTFETools() {
 		r.mcpServer.AddTool(tool.Tool, tool.Handler)
 	}
 
+	if toolsets.IsToolEnabled("create_project", r.enabledToolsets) {
+		tool := r.createDynamicTFETool("create_project", tfeTools.CreateProject)
+		r.mcpServer.AddTool(tool.Tool, tool.Handler)
+	}
+
+	if isTerraformOperationsEnabled() && toolsets.IsToolEnabled("delete_project", r.enabledToolsets) {
+		tool := r.createDynamicTFETool("delete_project", tfeTools.DeleteProject)
+		r.mcpServer.AddTool(tool.Tool, tool.Handler)
+	}
+
+	// Terraform toolset - Team tools
+	if toolsets.IsToolEnabled("create_team", r.enabledToolsets) {
+		tool := r.createDynamicTFETool("create_team", tfeTools.CreateTeam)
+		r.mcpServer.AddTool(tool.Tool, tool.Handler)
+	}
+
 	// Terraform toolset - Workspace management tools
 	if toolsets.IsToolEnabled("list_workspaces", r.enabledToolsets) {
 		tool := r.createDynamicTFETool("list_workspaces", tfeTools.ListWorkspaces)
@@ -137,24 +153,11 @@ func (r *DynamicToolRegistry) registerTFETools() {
 		r.mcpServer.AddTool(tool.Tool, tool.Handler)
 	}
 
-	// Only register create_project if TF operations are enabled AND toolset is enabled
-	if isTerraformOperationsEnabled() && toolsets.IsToolEnabled("create_project", r.enabledToolsets) {
-		tool := r.createDynamicTFETool("create_project", tfeTools.CreateProject)
-		r.mcpServer.AddTool(tool.Tool, tool.Handler)
-	}
-
-	// Only register delete_project if TF operations are enabled AND toolset is enabled
-	if isTerraformOperationsEnabled() && toolsets.IsToolEnabled("delete_project", r.enabledToolsets) {
-		tool := r.createDynamicTFETool("delete_project", tfeTools.DeleteProject)
-		r.mcpServer.AddTool(tool.Tool, tool.Handler)
-	}
-
-	// Only register delete_workspace_safely if TF operations are enabled AND toolset is enabled
 	if isTerraformOperationsEnabled() && toolsets.IsToolEnabled("delete_workspace_safely", r.enabledToolsets) {
 		tool := r.createDynamicTFETool("delete_workspace_safely", tfeTools.DeleteWorkspaceSafely)
 		r.mcpServer.AddTool(tool.Tool, tool.Handler)
 	}
-	// Only register force_unlock_workspace if TF operations are enabled AND toolset is enabled
+
 	if isTerraformOperationsEnabled() && toolsets.IsToolEnabled("force_unlock_workspace", r.enabledToolsets) {
 		tool := r.createDynamicTFETool("force_unlock_workspace", tfeTools.ForceUnlockWorkspace)
 		r.mcpServer.AddTool(tool.Tool, tool.Handler)
--- a/pkg/tools/tfe/create_team.go
+++ b/pkg/tools/tfe/create_team.go
@@ -1,0 +1,101 @@
+// Copyright IBM Corp. 2025
+// SPDX-License-Identifier: MPL-2.0
+
+package tools
+
+import (
+	"context"
+	"encoding/json"
+	"slices"
+	"strings"
+
+	"github.com/hashicorp/go-tfe"
+	"github.com/hashicorp/terraform-mcp-server/pkg/client"
+	log "github.com/sirupsen/logrus"
+
+	"github.com/mark3labs/mcp-go/mcp"
+	"github.com/mark3labs/mcp-go/server"
+)
+
+var validTeamVisibilities = []string{"secret", "organization"}
+var validTeamVisibilitiesStr = strings.Join(validTeamVisibilities, ", ")
+
+type TeamSummary struct {
+	ID         string `json:"team_id"`
+	Name       string `json:"team_name"`
+	Visibility string `json:"visibility"`
+}
+
+// CreateTeam creates a tool to create a new team in a Terraform Cloud/Enterprise organization.
+func CreateTeam(logger *log.Logger) server.ServerTool {
+	return server.ServerTool{
+		Tool: mcp.NewTool("create_team",
+			mcp.WithDescription(`Creates a new team in a Terraform Cloud/Enterprise organization.`),
+			mcp.WithTitleAnnotation("Create a new team in a Terraform organization"),
+			mcp.WithOpenWorldHintAnnotation(true),
+			mcp.WithReadOnlyHintAnnotation(false),
+			mcp.WithDestructiveHintAnnotation(false),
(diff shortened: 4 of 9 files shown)


pkg/tools/tfe/create_team.go
```diff
@@ -0,0 +1,101 @@
...
+)
+
+var validTeamVisibilities = []string{"secret", "organization"}
+var validTeamVisibilitiesStr = strings.Join(validTeamVisibilities, ", ")
+
+type TeamSummary struct {
+	ID         string `json:"team_id"`
+	Name       string `json:"team_name"`
+	Visibility string `json:"visibility"`
+}
```
Marin2409, line 28: It looks like `TeamSummary` is missing the optional `UserCount` field.

Based on this [ticket](https://hashicorp.atlassian.net/browse/TFECO-12908), the response should include `UserCount` when available:

> Returns: created team summary — ID, Name, Visibility (and optionally `UserCount`)

The TFE SDK's `Team` struct already exposes `UserCount` (`users-count`), and it is populated on the object returned by `tfeClient.Teams.Create(...)`.

This is how the updated struct would look like:

```go
// Current
type TeamSummary struct {
    ID         string `json:"team_id"`
    Name       string `json:"team_name"`
    Visibility string `json:"visibility"`
}

// Suggested
type TeamSummary struct {
    ID         string `json:"team_id"`
    Name       string `json:"team_name"`
    Visibility string `json:"visibility"`
    UserCount  int    `json:"user_count,omitempty"`
}
```

Something else worth mentioning, using `omitempty` would omit the field when its value is zero, which keeps the response concise for newly created teams that have no members yet.

pkg/tools/tfe/create_team.go
```diff
@@ -0,0 +1,101 @@
...
+	if visibility != "" {
+		options.Visibility = tfe.String(visibility)
+	}
+
+	team, err := tfeClient.Teams.Create(ctx, terraformOrgName, options)
+	if err != nil {
+		return ToolErrorf(logger, "failed to create team %q in org %q: %v", teamName, terraformOrgName, err)
+	}
+
+	teamJSON, err := json.Marshal(&TeamSummary{team.ID, team.Name, team.Visibility})
```
Marin2409: Then here, update this call to pass `team.UserCount` as the fourth field when constructing `TeamSummary`.

```go
// Current
teamJSON, err := json.Marshal(&TeamSummary{
    team.ID,
    team.Name,
    team.Visibility,
})

// Suggested
teamJSON, err := json.Marshal(&TeamSummary{
    team.ID,
    team.Name,
    team.Visibility,
    team.UserCount,
})
```

This ensures the `UserCount` field is populated in the JSON response when it is available from the TFE SDK.

pkg/tools/tfe/create_team.go
```diff
@@ -0,0 +1,101 @@
...
+				mcp.Description(terraformOrgNameDescription),
+			),
+			mcp.WithString("team_name",
+				mcp.Required(),
+				mcp.Description("The unique name of the team to create in the Terraform organization"),
+			),
+			mcp.WithString("visibility",
+				mcp.Description("Optional team visibility: "+validTeamVisibilitiesStr+`. Defaults to "secret" if not set.`),
+			),
+		),
```
Marin2409, line 50: Since we already know that the API default is to be "secret", I'd update the description to clarify about where it comes from, maybe like this: 

```go
mcp.Description("Optional team visibility: " + validTeamVisibilitiesStr + `. If omitted, the API defaults to "secret".`)
```

It currently says:

> `Defaults to "secret" if not set.`

The handler doesn't actually set a default value. If `visibility` is omitted, it passes an empty value to the TFE API, and the API itself applies the `"secret"` default.

This change makes it clear that the default is applied by the TFE API rather than by the tool itself, and it avoids the tool duplicating or overriding the API's default behavior if it ever changes in the future.

You could also make the tool the one who sets the field to "secret" as default:

```go
visibility := strings.ToLower(strings.TrimSpace(request.GetString("visibility", "secret")))
```
But then if the API changes we would have to come back and change it again. So I wouldn't recommend it. 

Image

pkg/tools/dynamic_tool.go
```diff
@@ -116,6 +116,22 @@ func (r *DynamicToolRegistry) registerTFETools() {
...
+	if isTerraformOperationsEnabled() && toolsets.IsToolEnabled("delete_project", r.enabledToolsets) {
+		tool := r.createDynamicTFETool("delete_project", tfeTools.DeleteProject)
+		r.mcpServer.AddTool(tool.Tool, tool.Handler)
+	}
+
+	// Terraform toolset - Team tools
+	if toolsets.IsToolEnabled("create_team", r.enabledToolsets) {
+		tool := r.createDynamicTFETool("create_team", tfeTools.CreateTeam)
+		r.mcpServer.AddTool(tool.Tool, tool.Handler)
+	}
```
Marin2409, line 133: Remember that `create_team` is a write operation and needs the `isTerraformOperationsEnabled()` gate. Without it, the tool registers and is callable even when Terraform write operations are supposed to be disabled (e.g. `ENABLE_TF_OPERATIONS=false`)

Here, you can simply add the `isTerraformOperationsEnabled()` condition before the `toolsets.IsToolEnabled`, like this:

```go
if isTerraformOperationsEnabled() && toolsets.IsToolEnabled("create_team", r.enabledToolsets) {
// ...
}
```

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ FEATURES
 
 * [New Tool] `create_project` Creates a new Terraform project in the specified organization. Requires `terraform_org_name` and `project_name`.
 * [New Tool] `delete_project` Deletes a Terraform project by ID. Requires `project_id`. TFC/TFE will refuse to delete a project that still contains workspaces or stacks.
+* [New Tool] `create_team` Creates a new team in a Terraform Cloud/Enterprise organization. Requires `terraform_org_name` and `team_name`; optional `visibility` ("secret" or "organization").
```
Marin2409: This branch currently has a merge conflict in `CHANGELOG.md`. Could you pull the latest changes from main, resolve the conflict, and update the branch?

Marin2409 (commented):
Hi @glbyktjys, I left some comments on this PR. The branch has a merge conflict in `CHANGELOG.md`. Could you pull the latest changes from main, resolve the conflict, and update the branch?

pkg/tools/dynamic_tool.go
jrhouston: I think it's OK for this tool not to be gated. The `ENABLE_TF_OPERATIONS` var is really for gating potentially destructive operations, i.e. deleting stuff and triggering runs

We can probably rename the var to `PREVENT_DESTRUCTIVE_OPERATIONS` or something like that in a future release.

pkg/tools/tfe/create_team.go
glbyktjys: Good catch, thanks! I originally left UserCount out because a newly created team generally has no members, and I thought the count might be more useful in a future team-membership management tool. However, since it is already returned by the SDK, I’ll add it to the summary here.

glbyktjys:
@jrhouston @Marin2409 Ready for another review when you have a chance. Thanks!

Marin2409 (commented):
This PR looks good, just make sure to update your branch with latest main and then it will be good for approval. Thanks @glbyktjys !!!

jrhouston:
@glbyktjys I think this is good to merge. Can you rebase it?

glbyktjys:
> @glbyktjys I think this is good to merge. Can you rebase it?

Thanks! I’ll rebase it and add another test that calls the go-tfe API directly before merging.

test/terraform/team_test.go
```diff
@@ -0,0 +1,59 @@
+package terraform
```
jrhouston, line 1: Let's have a `teams_test.go` instead and all the tests relating to teams can go in this file.

test/terraform/terraform_test.go
```diff
@@ -34,6 +37,8 @@ func (t *authTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 func init() {
 	mcpEndpoint = os.Getenv("TF_MCP_ENDPOINT")
 	tfeToken = os.Getenv("TFE_TOKEN")
+	tfeAddress = os.Getenv("TFE_ADDRESS")
+	tfeOrgName = os.Getenv("TFE_ORG_NAME")
```
jrhouston, line 55: @Marin2409 is also doing this in one of his PRs – coordinate on the naming here so we don't accidentally duplicate this

test/terraform/create_team_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+	if tfeOrgName == "" {
+		t.Skip("You need to supply TFE_ORG_NAME to run this test")
+	}
+
+	s := newTestingSession(t)
+	defer s.Close()
+
+	client := tfeClient(t)
+	teamName := fmt.Sprintf("mcp-test-team-%d", time.Now().Unix())
```
jrhouston: I'd prefer we didn't use the UNIX timestamp as a random suffix to make the names unique. It's temping because it seems unlikely that it will collide – because how likely it is it that two people will open a PR or run the tests at the exact same second? Later we might parallelize the test runs with t.Parallel(), establishing this pattern will make collision more likely than it seems. 

The right tool here is either:

1. Generate a UUID as the suffix, using [google/uuid](https://github.com/google/uuid). Most durable way to have unique names, but it does make the names really long and ugly. Not sure if we'll hit any length limits doing it this way. 
2. Write a helper function that generates a random alphanumeric suffix of N length. This has been done a bunch of different ways across hashi codebases, here's an example in the AWS provider: https://github.com/hashicorp/terraform-provider-aws/blob/af74ba133fa4a45144363afa25329a843a73257c/internal/acctest/vcr.go#L578-L604

test/terraform/team_test.go
```diff
@@ -0,0 +1,59 @@
...
+		"team_name":          teamName,
+		"visibility":         visibility,
+	})
+
+	require.False(t, result.IsError, "create_team returned an error: %s", resultText)
+	require.NotEmpty(t, resultText, "Tool call result must not be empty")
+
+	teamID := gjson.Get(resultText, "team_id").String()
+	require.NotEmpty(t, teamID, "Tool response should include a team ID")
+	defer client.Teams.Delete(t.Context(), teamID)
```
jrhouston, line 43: using defer to clean this up was the right choice in case the test fails, thanks 😄

test/terraform/team_test.go
```diff
@@ -0,0 +1,59 @@
...
+		"Tool response should report the created team name")
+
+	assert.Equal(t, visibility, gjson.Get(resultText, "visibility").String(),
+		"Tool response should report the requested visibility")
+
+	// Verify against the TFE API directly
+	createdTeam, err := client.Teams.Read(t.Context(), teamID)
+	require.NoError(t, err, "Team reported as created should be readable via the TFE API")
+	assert.Equal(t, teamName, createdTeam.Name, "Team name in TFE should match what was requested")
+	assert.Equal(t, visibility, createdTeam.Visibility)
```
jrhouston, line 55: Yeah this is a good idea because the tool doesn't actually return this – it just tells you the ID.

test/terraform/create_team_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+	assert.Equal(t, teamName, gjson.Get(resultText, "team_name").String(),
+		"Tool response should report the created team name")
+
+	assert.Equal(t, visibility, gjson.Get(resultText, "visibility").String(),
+		"Tool response should report the requested visibility")
+
+	// Verify against the TFE API directly
+	createdTeam, err := client.Teams.Read(t.Context(), teamID)
+	require.NoError(t, err, "Team reported as created should be readable via the TFE API")
```
jrhouston:
```suggestion
	require.NoError(t, err, "Team reported as created but produced an error when reading")
```

test/terraform/create_team_test.go
```diff
@@ -0,0 +1,59 @@
...
+	assert.Equal(t, teamName, createdTeam.Name, "Team name in TFE should match what was requested")
```
jrhouston:
```suggestion
	assert.Equal(t, teamName, createdTeam.Name, "Created team name does not match requested name")
```

jrhouston (changes requested):

test/terraform/terraform_test.go
glbyktjys: @Marin2409 Shall we go with `tfeOrgName` to make it consistent with `tfeToken` above?

Marin2409: Yes, that's what I have for list_teams tool [PR](https://github.com/hashicorp/terraform-mcp-server/pull/431).

You'll see there how I'm handling vars

Marin2409: Im already testing for `list` so just make the ones for `create`.

.github/workflows/hcpt_test.yml
```diff
@@ -35,5 +35,6 @@ jobs:
         env:
           TF_MCP_ENDPOINT: http://localhost:8080/mcp
           TFE_TOKEN: ${{ secrets.TFE_TOKEN_TESTING }}
+          TFE_ORG_NAME: ${{ secrets.TEST_TF_ORG_NAME }}
```
Marin2409: Instead of `secrets` do `vars` -> `vars.TEST_TF_ORG_NAME`

glbyktjys: Thank you!

test/terraform/team_test.go
```diff
@@ -0,0 +1,71 @@
...
+func randomTeamName() string {
+	const suffixLength = 8
+
+	suffix := make([]byte, suffixLength)
+	for i := range suffix {
+		suffix[i] = alphaNum[rand.Intn(len(alphaNum))]
+	}
+
+	return "mcp-test-team-" + string(suffix)
+}
```
jrhouston: We're gonna do this in more than one spot I bet. 

Maybe call this func `randomName()` and have it accept a prefix so you can do like `randomName("team-")` `randomName("project-")` etc. 

edit: also move the helper func into `terraform_test.go`

jrhouston (changes requested):
One last thing 😄

test/terraform/team_test.go
glbyktjys: Sounds good! Updated accordingly.

jrhouston (approved):
Awesome, great job @glbyktjys 🚀

PR #429, opened 2026-07-29
Add test harness for testing TFE tools using MCP SDK

jrhouston:
This PR adds a test harness using the official Go SDK for MCP to create a client and test TFE tool calls against a real server.

## PCI review checklist


test/terraform/organizations_test.go
```diff
@@ -0,0 +1,32 @@
...
+
+	"github.com/modelcontextprotocol/go-sdk/mcp"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/tidwall/gjson"
+)
+
+func TestListOrganziations(t *testing.T) {
+	s := newTestingSession(t)
+	defer s.Close()
```
glbyktjys, line 13: Do we need to add context.WithTimeout() for the connection in case anything happened and hanging process?

jrhouston: Yeah good call – I'll have it timeout after a few seconds.

test/terraform/organizations_test.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/tidwall/gjson"
+)
+
+func TestListOrganziations(t *testing.T) {
```
glbyktjys: nit: TestListOrganizations

test/terraform/terraform_test.go
```diff
@@ -0,0 +1,109 @@
...
+func (t *authTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	req = req.Clone(req.Context())
+	req.Header.Set("Authorization", "Bearer "+t.token)
+	return t.roundtripper.RoundTrip(req)
+}
+
+func init() {
+	mcpEndpoint = os.Getenv("TF_MCP_ENDPOINT")
+	if mcpEndpoint == "" {
+		mcpEndpoint = "http://localhost:8080/mcp"
```
glbyktjys: Not entirely sure if I understand it correctly. It looks like this fallback cannot currently be reached. init() already assigns http://localhost:8080/mcp when TF_MCP_ENDPOINT is empty, so mcpEndpoint will no longer be empty when newTestingSession() runs?

jrhouston: yep you're right this logic is duplicated

glbyktjys (approved):
LGTM! Thanks for addressing the comments and adding the workflow.

jaylonmcshan19-x (approved):

PR #430, opened 2026-07-30
add organization allowlist tool middleware

glbyktjys:
## Description

Adds an organization allowlist middleware for tool calls made through the HTTP server.

When an allowlist is configured, the middleware checks tool requests that include the `terraform_org_name` argument. Requests are rejected if the specified organization is not included in the configured allowlist.

Tools that do not use `terraform_org_name` continue to run normally. If the allowlist is empty, the middleware is not initialized and all tool requests are passed through unchanged.

## Changes

- `org_allowlist_tool_middleware.go`: Adds `OrganizationAllowlistToolMiddleware`. Builds a lookup set from the configured organization allowlist. Allows tools without a `terraform_org_name` argument to bypass the allowlist check. Rejects tool calls when the specified organization is not in the allowlist. Logs rejected tool calls with the tool and organization names.
- `org_allowlist_tool_middleware_test.go`: Adds unit tests covering allowed organizations, rejected organizations, tools without an organization argument, and empty allowlists.
- `main.go`: Registers `OrganizationAllowlistToolMiddleware` in `runHTTPServer`.

## Testing

- Added unit tests for the organization allowlist middleware.
- Tested locally with MCP Inspector, including allowed, rejected, bypassed, and empty-allowlist cases.

### DIFF
--- a/pkg/client/org_allowlist_tool_middleware.go
+++ b/pkg/client/org_allowlist_tool_middleware.go
@@ -28,7 +28,7 @@ func OrganizationAllowlistToolMiddleware(allowlist []string, logger *log.Logger)
 	return func(nextToolHandler server.ToolHandlerFunc) server.ToolHandlerFunc {
 		return func(ctx context.Context, request mcp.CallToolRequest) (*mcp.CallToolResult, error) {
 			// if tool doesn't have orgNameArgument, skipped without need to check
-			organizationName := request.GetString(orgNameArgument, "")
+			organizationName := strings.TrimSpace(request.GetString(orgNameArgument, ""))
 			if organizationName == "" {
 				return nextToolHandler(ctx, request)
 			}
--- a/pkg/client/org_allowlist_tool_middleware.go
+++ b/pkg/client/org_allowlist_tool_middleware.go
@@ -16,24 +16,17 @@ import (
 const orgNameArgument = "terraform_org_name"
 
 func OrganizationAllowlistToolMiddleware(allowlist []string, logger *log.Logger) server.ToolHandlerMiddleware {
-	allowedOrganizations := OrganizationAllowlistSet(allowlist)
-
-	// if no allowed organization listed, skip to next toolmiddleware without need to initialize OrganizationAllowlistToolMiddleware
-	if len(allowedOrganizations) == 0 {
-		return func(nextToolHandler server.ToolHandlerFunc) server.ToolHandlerFunc {
-			return nextToolHandler
-		}
-	}
+	allowedOrganizations := buildAllowedOrganizationsMap(allowlist)
 
 	return func(nextToolHandler server.ToolHandlerFunc) server.ToolHandlerFunc {
 		return func(ctx context.Context, request mcp.CallToolRequest) (*mcp.CallToolResult, error) {
-			// if tool doesn't have orgNameArgument, skipped without need to check
-			organizationName := strings.TrimSpace(request.GetString(orgNameArgument, ""))
+			// Skip allowlist enforcement for tools without an organization argument.
+			organizationName := strings.ToLower(strings.TrimSpace(request.GetString(orgNameArgument, "")))
 			if organizationName == "" {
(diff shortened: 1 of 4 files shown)


pkg/client/org_allowlist_tool_middleware.go
```diff
@@ -0,0 +1,57 @@
...
+		}
+	}
+}
+
+// builds a lookup set from allowlist
+func OrganizationAllowlistSet(allowlist []string) map[string]struct{} {
+	allowedSet := make(map[string]struct{}, len(allowlist))
+	for _, organizationName := range allowlist {
+		if organizationName != "" {
+			allowedSet[organizationName] = struct{}{}
```
jrhouston: ~I think you want a `strings.ToLower(organizationName)` here too~

edit: nvm, `ParseOrganizationAllowlistCSV()` does it.

pkg/client/org_allowlist_tool_middleware.go
```diff
@@ -0,0 +1,57 @@
...
+
+func OrganizationAllowlistToolMiddleware(allowlist []string, logger *log.Logger) server.ToolHandlerMiddleware {
+	allowedOrganizations := OrganizationAllowlistSet(allowlist)
+
+	// if no allowed organization listed, skip to next toolmiddleware without need to initialize OrganizationAllowlistToolMiddleware
+	if len(allowedOrganizations) == 0 {
+		return func(nextToolHandler server.ToolHandlerFunc) server.ToolHandlerFunc {
+			return nextToolHandler
+		}
+	}
```
jrhouston: Instead of putting this logic here, you can put it here: https://github.com/hashicorp/terraform-mcp-server/pull/430/changes#diff-8c13ae9f2942599ce56f3e326c28b6172ce04a81a6d9c1e2f48a6cabdac3ae8fR48-R56

Break the `ServerOption` params into a new slice, and supply it to `NewServer` using the `...` operator, i.e

```golang
serverOpts := []server.ServerOption{
    server.WithHooks(hooks),
}
if len(organizationAllowList) > 0 {
    serverOpts = append(serverOpts, server.WithToolHandlerMiddleware(
	    client.OrganizationAllowlistToolMiddleware(organizationAllowlist, logger),
	),
}
hcServer, rateLimiter := NewServer(
    version.Version,
	logger,
	enabledToolsets,
	serverOpts...
)
```

That way, the middleware doesn't even get added if organization allowlist isn't present.

pkg/client/org_allowlist_tool_middleware.go
```diff
@@ -0,0 +1,57 @@
...
+					"Terraform organization %q is not allowed by this server", organizationName)), nil
+			}
+
+			return nextToolHandler(ctx, request)
+		}
+	}
+}
+
+// builds a lookup set from allowlist
+func OrganizationAllowlistSet(allowlist []string) map[string]struct{} {
```
jrhouston: On the naming of this function, calling it `SomethingSomethingSet` makes me think that it _sets_ something, when what you are actually doing is building a map with the implicit intention to use it like a set. 

I would call this function `buildAllowedOrganizationsMap` since you're not actually introducing a `Set` abstraction anywhere.

jrhouston (changes requested):

pkg/client/org_allowlist_tool_middleware.go
glbyktjys: yeah that make sense so it doesn't even need to be initialized when starting the HTTP server.

glbyktjys:
Tested locally using the HTTP server. Disallowed organizations are rejected before the tool handler executes, while the configured organization is allowed through.

I also found that mixed-case or whitespace-padded input, such as `"     terraform-Org"`, exposes an existing normalization issue in the `list_terraform_projects` tool handler. This is outside the scope of this PR, so I’ll address them in another PR by adding `strings.TrimSpace()` and any other necessary normalization in the relevant tool handlers.

jrhouston (approved):

PR #436, opened 2026-07-31
Add per-tool-call Instana tracing

jaylonmcshan19-x:
## PCI review checklist

This PR adds a span for each MCP tool call so we actually get traces for them in Instana.

currently the http handler is already wrapped with Instana's TracingHandlerFunc, but that
only traces at the http request level. With streamable-http the tool calls all go
over one long lived connection, so it's only really seeing the connection and not
the individual tool calls happening inside it. I believe that's why the service shows up but
there's no actual trace data for the tools.

in regards to testing > 

I tested locally and confirmed the hooks fire and a span gets created + finished for
each tool call. But I couldn't fully confirm the spans actually landing in Instana from my
local setup though unfortunately. ran into some issues with the  go-sensor announce handshake to my local dockerized agent kept failing due to weird networking issues, which I know its always a pain with docker on mac

### DIFF
--- a/cmd/terraform-mcp-server/main.go
+++ b/cmd/terraform-mcp-server/main.go
@@ -18,6 +18,8 @@ import (
 	"github.com/hashicorp/terraform-mcp-server/pkg/client"
 	"github.com/hashicorp/terraform-mcp-server/pkg/toolsets"
 	"github.com/hashicorp/terraform-mcp-server/version"
+	instana "github.com/instana/go-sensor"
+	ot "github.com/opentracing/opentracing-go"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/exporters/otlp/otlpmetric/otlpmetrichttp"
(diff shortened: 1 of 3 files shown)


cmd/terraform-mcp-server/main.go
```diff
@@ -88,8 +77,50 @@ func runHTTPServer(logger *log.Logger, host string, port string, endpointPath st
...
 	attachMetricsHooks(hooks, metricsConfig, logger)
+	attachInstanaTracingHooks(hooks, instanaCollector)
+
+	return streamableHTTPServerInit(ctx, hcServer, logger, host, port, endpointPath, heartbeatInterval, organizationAllowlist, instanaCollector)
+
+}
+
+// toolCallSpans tracks in-flight spans by request id so we can start one in
+// BeforeCallTool and close out the matching span in AfterCallTool.
+var toolCallSpans sync.Map
```
Shweta-Deshpande: Should this be local to the attachInstanaTracingHooks function?

Shweta-Deshpande: Should this be local to the attachInstanaTracingHooks function?

jaylonmcshan19-x: Fixed in new commit

Shweta-Deshpande (dismissed):
Please fix the merge conflicts

PR #445, opened 2026-08-04
add authentication bearer token to loadtest

glbyktjys:
## Description
Updates the loadtest client to support authenticated MCP requests using `TFE_TOKEN`. When `TFE_TOKEN` is set, the load test configures a custom HTTP transport that adds: `Authorization: Bearer <TFE_TOKEN>` to outgoing MCP requests. If the token is not set, the loadtest continues using the default unauthenticated HTTP client.

no-changelog-needed


jrhouston (approved):
Thanks for adding this! @glbyktjys

PR #448, opened 2026-08-05
set service.instance.id and k8s.pod.uid on OTel metrics resource

jaylonmcshan19-x:
## PCI review checklist

This PR adds `service.instance.id `to the OTel metrics resource, and `k8s.pod.uid` when we're
running in k8s. 

right now `resource.NewSchemaless` only sets `service.name` and `service.version`, so every
replica sends an identical resource and Instana collapses them into a single entity.
in prod all 6 pods show up as one entity id. we can't tell replicas apart, and the entity never links to a k8s pod,
which is why none of the `kubernetes.*` tags are available when attempting to build dashboard widgets.

### DIFF
--- a/pkg/client/metrics.go
+++ b/pkg/client/metrics.go
@@ -19,6 +19,8 @@ type MetricsConfig struct {
 	ExportInterval        time.Duration            // Controls the frequency of metric flushes
 	ServiceName           string                   // ServiceName identifies the source of the metrics (e.g., "terraform-mcp-server")
 	ServiceVersion        string                   // ServiceVersion helps track metrics across different deployments
+	ServiceInstanceID     string                   // ServiceInstanceID uniquely identifies this running instance, so replicas don't collapse into one entity
+	K8sPodUID             string                   // K8sPodUID links the instance to its Kubernetes pod. Empty when not running in Kubernetes
 	MeterProvider         *sdkmetric.MeterProvider // MeterProvider is the OTel provider used to create instruments
 	Attributes            []attribute.KeyValue     // Attributes are global labels applied to every metric emitted
 	EnableRuntimeMetrics  bool                     // EnableRuntimeMetrics toggles the collection of Go runtime stats (GC, Memory)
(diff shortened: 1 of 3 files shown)


pkg/client/metrics.go
```diff
@@ -18,6 +18,8 @@ type MetricsConfig struct {
 	ExportInterval        time.Duration            // Controls the frequency of metric flushes
 	ServiceName           string                   // ServiceName identifies the source of the metrics (e.g., "terraform-mcp-server")
 	ServiceVersion        string                   // ServiceVersion helps track metrics across different deployments
+	ServiceInstanceID     string                   // ServiceInstanceID uniquely identifies this running instance, so replicas don't collapse into one entity
+	K8sPodUID             string                   // K8sPodUID links the instance to its Kubernetes pod. Empty when not running in Kubernetes
```
jrhouston: Each of these does effectively the same thing: uniquely identify a replica of the server.

I don't think we need to be K8s specific with this. We can just have one env var `OTEL_INSTANCE_ID` which when running in K8s can be populated like this:

https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/

jrhouston (changes requested):

pkg/client/metrics.go
jaylonmcshan19-x: @jrhouston I agree on the two config fields, dropping K8sPodUID for a single OTEL_INSTANCE_ID!.

But I think we may want to still keep emitting both attributes though. instana's docs treat them as two separate links, spans to the otel entity via `service.name + service.instance.id`, then the entity to the pod via `k8s.pod.uid`. they call that one out as critical for pod linking in k8s. Also their reference collector config sets both and copies `k8s.pod.uid` into `service.instance.id`, so same value in two attributes is the documented pattern.

I think we can go with this flow >  one env var, emit `k8s.pod.uid` off it too, gated on `KUBERNETES_SERVICE_HOST` so stdio users don't end up with a pod uid that's really just a hostname what do you think?

Docs for reference > https://www.ibm.com/docs/en/instana-observability?topic=opentelemetry-related-service-mappings-infrastructure-correlation

jrhouston: Sounds good!

jrhouston (dismissed):

jrhouston:
@jaylonmcshan19-x needs to be rebased

jaylonmcshan19-x:
@jrhouston Rebased!

PR #433, opened 2026-07-30
Add `get_project` tool

Marin2409:
## New Tool: `get_project`

Adds the `get_project` MCP tool to the terraform toolset. Fetches detailed information about a single Terraform project by its ID. Useful when you already have a `project_id` (e.g. from `list_terraform_projects`) and need the full details of that project.

### What changed

- `pkg/tools/tfe/get_project.go` — tool definition, handler, and `ProjectDetails` response struct
- `pkg/tools/tfe/get_project_test.go` — unit tests.
- `pkg/tools/dynamic_tool.go` — registers `get_project` in the terraform toolset.
- `pkg/toolsets/mapping.go` — adds `get_project` to the terraform toolset
- `CHANGELOG.md` — adds entry under `1.2.0`

### Inputs / Output

**Required:** `project_id` (e.g. `prj-abc123def456`)

**Returns:**
```json
{
  "project_id": "prj-abc123def456",
  "project_name": "my-project",
  "description": "...",
  "default_execution_mode": "remote",
  "is_unified": false,
  "auto_destroy_activity_duration": "14d",
  "organization_name": "my-org"
}
```

## PCI review checklist

- [ ] I have documented a clear reason for, and description of, the change I am making.

- [ ] If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request.

- [ ] If applicable, I've documented the impact of any changes to security controls.

  Examples of changes to security controls include using new access control methods, adding or removing logging pipelines, etc.

PR #449, opened 2026-08-05
test(tfe-tests): add workspace listing integration test

MulyaP:
This PR adds test case for `list_workspaces` tool for testing it against a real server.

## PCI review checklist

### DIFF
--- a/test/terraform/workspace_test.go
+++ b/test/terraform/workspace_test.go
@@ -1,0 +1,37 @@
+package terraform
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/tidwall/gjson"
+)
+
+func TestListWorkspace(t *testing.T) {
+	s := newTestingSession(t)
+	defer s.Close()
+
+	result, resultText := callTool(t, s, "list_terraform_orgs", map[string]any{})
+
+	require.False(t, result.IsError, "Organization tool call result should not be an error")
+	require.NotEmpty(t, resultText, "Organiation tool call result must not be empty")
+
+	assert.NotEqual(t, int(gjson.Get(resultText, "items.#").Int()), 0, "Organization tool call result should not contain an empty list")
+	assert.NotEmpty(t, gjson.Get(resultText, "items.0.organization_name").String(), "Tool call result should contain organization names")
+	assert.NotEmpty(t, gjson.Get(resultText, "items.0.organization_email").String(), "Tool call result should contain organization email addresses")
+
+	orgName := gjson.Get(resultText, "items.0.organization_name").String()
+
+	result, resultText = callTool(t, s, "list_workspaces",
+	map[string]any{
+		"terraform_org_name": orgName,
+	})
+
+	require.False(t, result.IsError, "Workspace tool call result should not be an error")
+	require.NotEmpty(t, resultText, "Workspace tool call result should not be empty")
+	
+	assert.NotEqual(t, int(gjson.Get(resultText, "items.#").Int()), 0, "Workspace tool call result should not contain an empty list")
+	assert.NotEmpty(t, gjson.Get(resultText, "items.0.id").String(), "First workspace should have an ID")
+	assert.NotEmpty(t, gjson.Get(resultText, "items.0.workspace_name").String(), "First workspace should have a name")
+}
--- a/test/terraform/workspace_test.go
+++ b/test/terraform/workspace_test.go
@@ -3,267 +3,352 @@ package terraform
 import (
 	"testing"
 
+	"github.com/modelcontextprotocol/go-sdk/mcp"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"github.com/tidwall/gjson"
 )
 
-func TestWorkspaceReadOnlyTools(t *testing.T) {
+func TestWorkspaceHappyPath(t *testing.T) {
+	if tfeOrgName == "" {
+		t.Skip("TFE_ORG_NAME is not set; skipping workspace tools tests")
+	}
+	
+	client := tfeClient(t)
 	s := newTestingSession(t)
 	defer s.Close()
 
-	result, resultText := callTool(t, s, "list_terraform_orgs", map[string]any{})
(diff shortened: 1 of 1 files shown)


Marin2409 (commented):
Thanks, @MulyaP, for adding the HCPT workspace test cases!

The goal of `workspace_test.go` is to cover all workspace-related tools. I see you've already added coverage for `list_workspaces`, which is a great start.

We also have the following workspace tools that need to be included:

`get_workspace_details`, `create_workspace`, `create_no_code_workspace`, `update_workspace`, `delete_workspace_safely`, `list_workspace_variables`, `create_workspace_variable`, `update_workspace_variable`, `create_workspace_tags`, `read_workspace_tags`, `list_workspace_policy_sets`, `force_unlock_workspace`, `list_state_versions`, and `get_state_version`.

The idea is to test and simulate each of these tools against an actual HCPT server. As a reference, you can take a look at [PR #433](https://github.com/hashicorp/terraform-mcp-server/pull/433), where I followed the same approach for multiple project-related tools.

Also, you'll see how I handle tools that are gated behind `ENABLE_TF_OPERATIONS`. These are mostly destructive operation tools, but I'd still recommend checking `dynamic_tool.go` since there are a few non-destructive tools that are also gated behind `ENABLE_TF_OPERATIONS`. That should help determine which test cases need to be conditionally skipped when the environment variable isn't set.

Since my PR hasn't been merged yet, feel free to focus on the workspace tools that aren't gated behind `ENABLE_TF_OPERATIONS`. I can take care of adding the gated test cases once my PR lands. Of course, if you'd rather implement them now, that's totally fine too. 

Thanks again for getting the workspace tool tests against a real HCPT server started.

Mulya-Patel:
Thanks, @Marin2409. Thanks for giving me further directions, I will get right to it. To avoid conflicts, I think I should steer clear of tools gated behind `ENABLE_TF_OPERATIONS` till your PR is merged. I will start adding other workspace related tools' test in this PR itself.

test/terraform/workspace_test.go
```diff
@@ -0,0 +1,269 @@
...
+
+		t.Run("Non-existent workspace ID returns an error", func(t *testing.T) {
+			svResult, _ := callTool(t, s, "get_state_version", map[string]any{
+				"workspace_id": "ws-doesnotexistxyz",
+			})
+
+			assert.True(t, svResult.IsError, "get_state_version should return an error for a non-existent workspace ID")
+		})
+	})
+}
```
jrhouston: I think we should put these in their own file `state_version_test.go` as this tool is probably going to evolve to be more complex.

test/terraform/workspace_test.go
```diff
@@ -0,0 +1,269 @@
...
+	result, resultText := callTool(t, s, "list_terraform_orgs", map[string]any{})
+
+	require.False(t, result.IsError, "Organization tool call result should not be an error")
+	require.NotEmpty(t, resultText, "Organiation tool call result must not be empty")
+
+	assert.NotEqual(t, int(gjson.Get(resultText, "items.#").Int()), 0, "Organization tool call result should not contain an empty list")
+	assert.NotEmpty(t, gjson.Get(resultText, "items.0.organization_name").String(), "Tool call result should contain organization names")
+	assert.NotEmpty(t, gjson.Get(resultText, "items.0.organization_email").String(), "Tool call result should contain organization email addresses")
+
+	firstOrgName := gjson.Get(resultText, "items.0.organization_name").String()
```
jrhouston: We already test this in organizations_test.go

I think we can just create a variable `TEST_ORG_NAME` and configure it in the github actions workflow.

test/terraform/workspace_test.go
```diff
@@ -0,0 +1,269 @@
...
+					"terraform_org_name": firstOrgName,
+					"workspace_name":     firstWorkspaceName,
+				})
+
+			require.False(t, listVarsResult.IsError, "list_workspace_variables should not return an error")
+			require.NotEmpty(t, listVarsResultText, "list_workspace_variables result should not be empty")
+
+			assert.True(t, gjson.Get(listVarsResultText, "data").IsArray(), "list_workspace_variables result should contain a data array")
+
+			// Check if workspace has variables, if not, then skip variable specific assert statements
```
jrhouston: I think if we want to test this, then we should explicitly create them. Having a test that sometimes skips depending on what's in the test env is kinda weird. 

We can use the go-tfe client to create anything we need to test against.

jrhouston (changes requested):

test/terraform/workspace_test.go
Mulya-Patel: I removed it for now, I think its best to work on that in a new PR

test/terraform/workspace_test.go
Mulya-Patel: Added a variable `TEST_ORG_NAME` which reads from `secrets.HCPT_TEST_ORG_NAME`

test/terraform/workspace_test.go
Mulya-Patel: I refactored all workspace related test cases to follow this pattern of create->test->delete. Also had a separate function to test some error paths. Used the go-tfe client for cleanup so a tool failure does not change the org state.

test/terraform/workspace_test.go
```diff
@@ -0,0 +1,379 @@
...
+	if tfeOrgName == "" {
+		t.Skip("TFE_ORG_NAME is not set; skipping workspace tools tests")
+	}
+
+	requireTfOperations(t)
+
+	s := newTestingSession(t)
+	defer s.Close()
+
+	wsName := "mcp-server-integration-test-ws"
```
jrhouston: For testing, we should generate unique names in case the tests get run in parallel. 

Jiayue is adding a helper for this here: https://github.com/hashicorp/terraform-mcp-server/pull/427/changes#diff-5128ad3eb3ebb639de3477b2e2671e25304b2600300d90e8b2b7db841aa4ee2eR14

Once merged you should rebase this branch and use it.

jrhouston:
@Mulya-Patel Jiayue's changes were merged so you can now rebase this and use her randomName() helper func.

jrhouston (approved):

PR #450, opened 2026-08-07
Bootstrap new server using the official modelcontextprotocol/go-sdk

Shweta-Deshpande:
Initialize the server using the official mcp go-sdk package and make sure it stores all the current security updates.
Add a sample tool list_workspaces to use this new server.

## PCI review checklist


pkg/mcp-official/server.go
```diff
@@ -0,0 +1,28 @@
...
+	if heartbeatInterval > 0 {
+		log.Infof("HTTP heartbeat enabled with interval: %v", heartbeatInterval)
+		svrOptions = &mcp.ServerOptions{
+			KeepAlive: heartbeatInterval,
+		}
+	}
+	svr := mcp.NewServer(
+		&mcp.Implementation{
+			Name:    "terraform-mcp-official",
+			Version: "1.0.0",
```
jrhouston: Should grab this from the `version` package.

pkg/mcp-official/helper.go
```diff
@@ -0,0 +1,129 @@
+package mcpofficial
```
jrhouston: We should try to avoid naming packages things like `helper.go` `util.go` etc. as that's considered a faux pas in Go.

Most of the functions in this file are for client related stuff, so perhaps `client.go` is a more suitable name?

See this section in the Go style guide: https://google.github.io/styleguide/go/best-practices.html#util-packages

pkg/mcp-official/workspace.go
```diff
@@ -0,0 +1,125 @@
...
+		client = value.(*tfe.Client)
+	}
+	var err error
+	if client == nil {
+		log.Printf("TFE client not found, creating a new one")
+		client, err = CreateTfeClientForSession(ctx, session.ID())
+		if err != nil {
+			return nil, nil, err
+		}
+	}
```
jrhouston: I think we will need to rethink this because the [new spec for MCP has gone stateless](https://blog.modelcontextprotocol.io/posts/2026-07-28/) and so sessions are going away. 

I think it might be better just to drop sessions completely in the migration.

pkg/mcp-official/workspace.go
```diff
@@ -0,0 +1,125 @@
...
+	// Required field
+	TerraformOrgName string `json:"terraform_org_name" jsonschema:"The Terraform organization name"`
+
+	// Optional fields (will be empty strings if not provided)
+	ProjectID    string `json:"project_id,omitempty" jsonschema:"Filter by project ID"`
+	SearchQuery  string `json:"search_query,omitempty" jsonschema:"Search term"`
+	Tags         string `json:"tags,omitempty" jsonschema:"Comma-separated tags"`
+	ExcludeTags  string `json:"exclude_tags,omitempty" jsonschema:"Tags to exclude"`
+	WildcardName string `json:"wildcard_name,omitempty" jsonschema:"Wildcard pattern"`
+}
```
jrhouston: I know this is just an example, but the tool is called `ListWorkspaces` so prob the args struct should also be called `ListWorkspacesArguments`

pkg/mcp-official/helper.go
```diff
@@ -0,0 +1,129 @@
...
+		terraformSkipTLSVerify, err := strconv.ParseBool(terraformSkipTLSVerifyStr)
+		if err == nil {
+			return terraformSkipTLSVerify
+		}
+	}
+	return false
+}
+
+// createHTTPClient initializes a retryable HTTP client
+func createHTTPClient(insecureSkipVerify bool) *http.Client {
```
jrhouston: This code is copied verbatim form here: https://github.com/hashicorp/terraform-mcp-server/blob/592a6e07dcc0d254ff46b939c18277e234c852cc/pkg/client/registry.go#L27-L63

We could just export that function and use it inside this package?

jrhouston (changes requested):
Left a couple comments

cmd/terraform-mcp-server/init.go
```diff
@@ -364,6 +366,12 @@ func streamableHTTPServerInit(ctx context.Context, hcServer *server.MCPServer, l
 	mux.Handle(endpointPath, streamableServer)
 	mux.Handle(endpointPath+"/", streamableServer)
 
+	// Create the official go-sdk streamable server
+	officialStreamableServer := getOfficialStreamableServer(ctx, heartbeatInterval, isStateless, tlsConfig, corsConfig, logger, organizationAllowlist, enabledToolsets)
+	// Handle the /mcp endpoint with the official go-sdk streamable server (with security wrapper)
+	mux.Handle(endpointPath+"/official", officialStreamableServer)
+	mux.Handle(endpointPath+"/official/", officialStreamableServer)
```
jrhouston: Let's gate this behind an environment variable for now, so it's not switched on by default.

Could call it `TF_X_OFFICIAL_SDK` as the `TF_X` prefix has been used in a lot of other projects for experimental stuff.

pkg/mcp-official/server.go
Shweta-Deshpande: Done

pkg/mcp-official/helper.go
Shweta-Deshpande: Done

pkg/mcp-official/workspace.go
Shweta-Deshpande: Done

pkg/mcp-official/workspace.go
Shweta-Deshpande: Done

pkg/mcp-official/helper.go
Shweta-Deshpande: Yeah I copied it over to avoid changing the other calls to the original function. Done

cmd/terraform-mcp-server/init.go
Shweta-Deshpande: Done

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ FEATURES
 * [New Tool] `list_teams` Lists all teams within a given Terraform Cloud organization. Requires `terraform_org_name`. Optionally filter by exact team names (`team_names`), or substring search (`search_query`). Supports pagination.
 * [New Tool] `get_team` Fetches full details for a single team by ID, including members, organization access permissions, and SSO settings. Requires `team_id`. [441](https://github.com/hashicorp/terraform-mcp-server/pull/441)
 
+IMPROVEMENTS
+
+* Migrate code to use the official mcp go-sdk
+
```
jrhouston: Let's leave this out of the CHANGELOG for now until it's actually done – this PR is just bootstrapping the new server.

README.md
```diff
@@ -61,6 +61,7 @@ automation and interaction capabilities for Infrastructure as Code (IaC) develop
 | `OTEL_METRICS_ENDPOINT` | URL of your OTel Collector or backend | `localhost:4318` |
 | `INSTANA_ENABLED` | Enable Instana instrumentation (metrics and HTTP request tracing) for the streamable-http server. Requires an Instana agent that is reachable by the server. | `false` |
 | `INSTANA_SERVICE_NAME` | If Instana instrumentation is enabled, the service name to use for the MCP server | `terraform-mcp-server` |
+| `TF_X_OFFICIAL_SDK_ENABLED` | Enable the server with the official mcp go-sdk | `false` |
```
jrhouston: Leave this off of the README.md – it's an internal thing just for us (and any OSS people paying close attention to the code)

pkg/mcp-official/server.go
```diff
@@ -0,0 +1,30 @@
...
+		}
+	}
+	svr := mcp.NewServer(
+		&mcp.Implementation{
+			Name:    "terraform-mcp-official",
+			Version: version.Version,
+		},
+		svrOptions,
+	)
+	logger.Info("Official mcp go-sdk server instance started at 127.0.0.1:8080/mcp/official..")
```
jrhouston: This log line says it listens on 127.0.01 but that's not actually what happens here – remove this line.

jrhouston (changes requested):
Couple cosmetic things then we can merge and iterate on this.

pkg/mcp-official/server.go
Shweta-Deshpande: Thanks! Good catch. Is it ok if I use these functions? 

```
// getHTTPPort returns the port from environment variables or default
func getHTTPPort() string {
	if port := os.Getenv("TRANSPORT_PORT"); port != "" {
		return port
	}
	return "8080"
}

// getHTTPHost returns the host from environment variables or default
func getHTTPHost() string {
	if host := os.Getenv("TRANSPORT_HOST"); host != "" {
		return host
	}
	return "127.0.0.1"
}
```

Was hoping to have a log statement when I start this server.

README.md
Shweta-Deshpande: Makes sense.

pkg/mcp-official/server.go
jrhouston: There is already a log message for the server here: https://github.com/hashicorp/terraform-mcp-server/blob/main/cmd/terraform-mcp-server/init.go#L426

We are just attaching a new route to that server.

And you already have a log message to show that the official server has been enabled: https://github.com/hashicorp/terraform-mcp-server/pull/450/changes#diff-7ab12dbce0f7c555935008ef04909236a05ec1e07afcd4cb1216c045bea5ccd6R371

CHANGELOG.md
Shweta-Deshpande: Should we maybe just say Bootstrap instead of migrate?

jrhouston (approved):

PR #451, opened 2026-08-07
feat: add get_team tool to fetch full team details by ID (#441)

### LINKED ISSUE #441: Add `get_team` — Fetch full details for a single team by ID
### Background

The `list_teams` tool returns a summary of all teams in an organization (ID, name, visibility, user count) and is intentionally kept minimal to avoid bloating responses for orgs with many teams.

However, there is currently no way to fetch the full details of a specific team — including its members, organization access permissions, organization memberships, and SSO settings — once a team ID has been identified.

### Proposed Tool: `get_team`

Add a new `get_team` MCP tool that fetches the complete details for a single team by its ID, backed by the [`tfe.Teams.Read`](https://pkg.go.dev/github.com/hashicorp/go-tfe#Teams) API endpoint (`GET /teams/:team_id`).

This complements the `list_teams` → `get_team` workflow:
1. Use `list_teams` to discover team IDs.
2. Use `get_team` with a specific `team_id` to fetch full team details.

### Inputs

| Parameter | Required | Description |
|---|---|---|
| `team_id` | YES | The ID of the team to retrieve (e.g. `team-abc123`). |

### Output

Full team details, including:
- `id`, `name`, `visibility`
- `users-count`
- `organization-access` — the team's permission flags on the organization (manage workspaces, manage policies, etc.)
- `permissions` — current token's permissions on the team (can-destroy, can-update-membership)
- `sso-team-id`
- `users` — full list of team members
- `organization-memberships`

### Relevant API

- TFC API docs: https://developer.hashicorp.com/terraform/cloud-docs/api-docs/teams#show-a-team
- go-tfe method: `tfe.Teams.Read(ctx, teamID)`
- go-tfe type: `tfe.Team`

Saikiran5604:
## Summary
This PR resolves #441 by introducing the new `get_team` MCP tool. It complements the existing `list_teams` workflow by allowing users to fetch full, detailed attributes for a specific team using its ID.

## Changes
- Added `GetTeam` tool definition and handler in `pkg/tools/tfe/get_team.go`, backed by the `go-tfe` SDK method `client.Teams.Read(ctx, teamID)`.
- Registered `get_team` dynamically in `pkg/tools/dynamic_tool.go`.
- Added comprehensive unit and parameter validation tests in `pkg/tools/tfe/get_team_test.go`.

## Workflow
1. Use `list_teams` to discover team IDs.
2. Use `get_team` with a specific `team_id` (e.g., `team-abc123`) to fetch complete membership, permission flags, and SSO settings.

Closes #441

## PCI review checklist

### DIFF
--- a/pkg/tools/tfe/get_team.go
+++ b/pkg/tools/tfe/get_team.go
@@ -4,10 +4,11 @@
 package tools
 
 import (
+	"bytes"
 	"context"
-	"encoding/json"
 	"strings"
 
+	"github.com/google/jsonapi"
 	"github.com/hashicorp/terraform-mcp-server/pkg/client"
 	log "github.com/sirupsen/logrus"
 
@@ -55,10 +56,10 @@ func getTeamHandler(ctx context.Context, request mcp.CallToolRequest, logger *lo
 		return ToolErrorf(logger, "failed to read team %q: %v", teamID, err)
 	}
 
-	teamJSON, err := json.Marshal(team)
-	if err != nil {
+	buf := bytes.NewBuffer(nil)
+	if err := jsonapi.MarshalPayloadWithoutIncluded(buf, team); err != nil {
 		return ToolError(logger, "failed to marshal team details", err)
 	}
 
-	return mcp.NewToolResultText(string(teamJSON)), nil
+	return mcp.NewToolResultText(buf.String()), nil
 }
--- a/pkg/tools/tfe/get_team.go
+++ b/pkg/tools/tfe/get_team.go
@@ -1,0 +1,64 @@
+// Copyright IBM Corp. 2025
+// SPDX-License-Identifier: MPL-2.0
+
+package tools
+
+import (
+	"context"
+	"encoding/json"
+	"strings"
+
+	"github.com/hashicorp/terraform-mcp-server/pkg/client"
+	log "github.com/sirupsen/logrus"
+
+	"github.com/mark3labs/mcp-go/mcp"
+	"github.com/mark3labs/mcp-go/server"
+)
+
+// GetTeam creates a tool to fetch full details for a single team by ID.
+func GetTeam(logger *log.Logger) server.ServerTool {
+	return server.ServerTool{
+		Tool: mcp.NewTool("get_team",
+			mcp.WithDescription(`Fetch full details for a single team by ID, including members, organization access permissions, and SSO settings.`),
+			mcp.WithTitleAnnotation("Fetch full details for a single team by ID"),
+			mcp.WithOpenWorldHintAnnotation(true),
+			mcp.WithReadOnlyHintAnnotation(true),
+			mcp.WithDestructiveHintAnnotation(false),
+			mcp.WithString("team_id",
+				mcp.Required(),
+				mcp.Description("The ID of the team to retrieve (e.g. team-abc123)."),
+			),
+		),
+		Handler: func(ctx context.Context, request mcp.CallToolRequest) (*mcp.CallToolResult, error) {
+			return getTeamHandler(ctx, request, logger)
+		},
+	}
+}
+
+func getTeamHandler(ctx context.Context, request mcp.CallToolRequest, logger *log.Logger) (*mcp.CallToolResult, error) {
+	teamID, err := request.RequireString("team_id")
+	if err != nil {
+		return ToolError(logger, "missing required input: team_id", err)
+	}
+	teamID = strings.TrimSpace(teamID)
+
+	tfeClient, err := client.GetTfeClientFromContext(ctx, logger)
+	if err != nil {
+		return ToolError(logger, "failed to get Terraform client", err)
+	}
+	if tfeClient == nil {
+		return ToolError(logger, "failed to get Terraform client - ensure TFE_TOKEN and TFE_ADDRESS are configured", nil)
+	}
+
+	team, err := tfeClient.Teams.Read(ctx, teamID)
+	if err != nil {
+		return ToolErrorf(logger, "failed to read team %q: %v", teamID, err)
+	}
+
+	teamJSON, err := json.Marshal(team)
+	if err != nil {
+		return ToolError(logger, "failed to marshal team details", err)
+	}
+
+	return mcp.NewToolResultText(string(teamJSON)), nil
+}
--- a/pkg/tools/tfe/get_team.go
+++ b/pkg/tools/tfe/get_team.go
@@ -8,7 +8,7 @@ import (
 	"context"
 	"strings"
 
-	"github.com/google/jsonapi"
+	"github.com/hashicorp/jsonapi"
 	"github.com/hashicorp/terraform-mcp-server/pkg/client"
 	log "github.com/sirupsen/logrus"
 
--- a/test/terraform/terraform_test.go
+++ b/test/terraform/terraform_test.go
@@ -11,6 +11,9 @@ import (
 
 	"github.com/hashicorp/go-tfe"
 	"github.com/modelcontextprotocol/go-sdk/mcp"
+	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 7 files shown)


pkg/tools/tfe/get_team.go
```diff
@@ -0,0 +1,64 @@
...
+	if tfeClient == nil {
+		return ToolError(logger, "failed to get Terraform client - ensure TFE_TOKEN and TFE_ADDRESS are configured", nil)
+	}
+
+	team, err := tfeClient.Teams.Read(ctx, teamID)
+	if err != nil {
+		return ToolErrorf(logger, "failed to read team %q: %v", teamID, err)
+	}
+
+	teamJSON, err := json.Marshal(team)
```
Marin2409: Could you swap `json.Marshal` for `jsonapi.MarshalPayloadWithoutIncluded`?

```go
buf := bytes.NewBuffer(nil)
err = jsonapi.MarshalPayloadWithoutIncluded(buf, team)
if err != nil {
    return ToolError(logger, "failed to marshal team details", err)
}
```

The reason is that `tfe.Team` fields are annotated exclusively with `jsonapi:` struct tags:

```go
ID         string `jsonapi:"primary,teams"`
Name       string `jsonapi:"attr,name"`
Visibility string `jsonapi:"attr,visibility"`
```

`encoding/json` is completely blind to `jsonapi:` tags, it only reads `json:` tags. Since none of `tfe.Team`'s fields have `json:` tags, `json.Marshal` falls back to the exported Go field names, which causes a couple of issues:

- **Incorrect field names** — the output will contain `"Name"` instead of `"name"`, `"UserCount"` instead of `"users-count"`, etc. These don't match the Terraform API schema.
- **Relations aren't serialized correctly** — fields like `Users []*User` and `OrganizationMemberships` are marshaled as raw nested Go structs instead of proper JSON:API resource linkages. Using `MarshalPayloadWithoutIncluded` also intentionally omits the `included` section, keeping the response concise and avoiding large nested payloads.

You can see the same pattern in `get_run_details.go` (around line 52), where `jsonapi.MarshalPayloadWithoutIncluded` is used for the same reason.

Thanks!

Marin2409 (changes requested):
Hey @Saikiran5604, thanks for picking this up! 🙌

Just a couple of things to add:

1. In `pkg/toolsets/mapping.go`, add `"get_team": Terraform` to the `ToolToToolset` map so the tool is available when the `terraform` toolset is enabled.

3. In `test/terraform/teams_test.go`, please add an integration test `TestGetTeam` following the same pattern as the existing `TestListTeams` and `TestCreateTeam` tests. It should use a real `team_id` from the API (for example, the owners team), invoke the tool, and validate the JSON:API response.

   ```go
   func TestGetTeam(t *testing.T) {
       client := tfeClient(t)
       entitlements, err := client.Organizations.ReadEntitlements(t.Context(), tfeOrgName)
       require.NoError(t, err, "Failed to read entitlements for organization %q", tfeOrgName)
       if !entitlements.Teams {
           t.Skipf("Organization %q does not have the Teams entitlement", tfeOrgName)
       }

       // Get a real team ID to look up (the owners team always exists)
       teams, err := client.Teams.List(t.Context(), tfeOrgName, nil)
       require.NoError(t, err)
       require.NotEmpty(t, teams.Items, "Expected at least one team in the organization")
       teamID := teams.Items[0].ID

       s := newTestingSession(t)
       defer s.Close()

       result, resultText := callTool(t, s, "get_team", map[string]any{
           "team_id": teamID,
       })

       require.False(t, result.IsError, "Tool call result should not be an error")
       require.NotEmpty(t, resultText, "Tool call result must not be empty")

       // MarshalPayloadWithoutIncluded returns a JSON:API envelope, not a flat object.
       assert.Equal(t, teamID, gjson.Get(resultText, "data.id").String(), "Response should contain the requested team ID")
       assert.NotEmpty(t, gjson.Get(resultText, "data.attributes.name").String(), "Response should contain the team name")
       assert.NotEmpty(t, gjson.Get(resultText, "data.attributes.visibility").String(), "Response should contain the team visibility")
       assert.True(t, gjson.Get(resultText, "data.attributes.users-count").Exists(), "Response should contain the user count field")
   }
   ```

The key thing to note is that `get_team` uses `jsonapi.MarshalPayloadWithoutIncluded`, unlike tools such as `create_team` or `list_teams`, which return custom response structs. Because of that, the response is wrapped in a JSON:API envelope, so the `gjson` paths should be `data.id` and `data.attributes.<field>` rather than the flat structure used by those other tools.

pkg/tools/tfe/get_team.go
Saikiran5604: Hi @Marin2409 , thank you for the detailed feedback and explanation regarding `jsonapi.MarshalPayloadWithoutIncluded`! That makes complete sense.

I have addressed all your points:
- Swapped standard `json.Marshal` for `jsonapi.MarshalPayloadWithoutIncluded` in `pkg/tools/tfe/get_team.go`.
- Added `"get_team": Terraform` to the `ToolToToolset` map in `pkg/toolsets/mapping.go`.
- Added the `TestGetTeam` integration test following your pattern in `test/terraform/teams_test.go`.

All changes have been pushed. Please let me know if everything looks good now! Thanks again for your guidance.

test/terraform/terraform_test.go
```diff
@@ -151,3 +154,34 @@ func callTool(t *testing.T, s *mcp.ClientSession, toolName string, arguments map
...
+
+	require.False(t, result.IsError, "Tool call result should not be an error")
+	require.NotEmpty(t, resultText, "Tool call result must not be empty")
+
+	// MarshalPayloadWithoutIncluded returns a JSON:API envelope, not a flat object.
+	assert.Equal(t, teamID, gjson.Get(resultText, "data.id").String(), "Response should contain the requested team ID")
+	assert.NotEmpty(t, gjson.Get(resultText, "data.attributes.name").String(), "Response should contain the team name")
+	assert.NotEmpty(t, gjson.Get(resultText, "data.attributes.visibility").String(), "Response should contain the team visibility")
+	assert.True(t, gjson.Get(resultText, "data.attributes.users-count").Exists(), "Response should contain the user count field")
+}
```
Marin2409: One small thing: `TestGetTeam` should live in `test/terraform/teams_test.go`, alongside the other team-related integration tests.

You can find `teams_test.go` under the `test/terraform/` directory, so just move the `TestGetTeam` function from `test/terraform/terraform_test.go` into `test/terraform/teams_test.go`.

pkg/tools/tfe/get_team.go
```diff
@@ -0,0 +1,65 @@
...
+// SPDX-License-Identifier: MPL-2.0
+
+package tools
+
+import (
+	"bytes"
+	"context"
+	"strings"
+
+	"github.com/google/jsonapi"
```
jrhouston: I think your goimports has added the wrong package here. 

You probably wanted: https://github.com/hashicorp/jsonapi

jrhouston (changes requested):

pkg/tools/tfe/get_team.go
Saikiran5604: Ack , Fixed it .
Thanks !

test/terraform/terraform_test.go
Saikiran5604: Ack ! Moved to TestGetTeam function to `test/terraform_test.go` .
Thanks !

Marin2409 (approved):
Looks great! Thank you for your contribution @Saikiran5604 :)

jrhouston:
This needs to be rebased then it's good to go

Saikiran5604:
@jrhouston  Could you review and approve the changes :)

jrhouston (approved):
Thanks for adding this one @Saikiran5604

PR #456, opened 2026-08-07
Fix for ID not returning in get_workspace_details tool

MulyaP:
This PR contains the fix for ID not getting returned in the response of `get_workspace_details` tool.

### Issue: 
ID field in `tfe.Workspace` type was not being populated because it is tagged as `jsonapi:"primary,workspaces"`  while the `tfe.Workspace` type is a struct nested inside `jsonapi:"attr"`. When the struct is nested inside `jsonapi:"attr"`, the JSON marshaller only processes the `attr`-tagged fields and silently ignores the `primary` tagged fields.

### Fix: 
Added a new field `WorkspaceID` in the `WorkspaceToolResponse` type and explicitly populated it with the ID that was returned as a response of `tfeClient.Workspaces.Read`.

## PCI review checklist

### DIFF
--- a/pkg/client/types.go
+++ b/pkg/client/types.go
@@ -440,11 +440,12 @@ type TerraformPolicyDetails struct {
 }
 
 type WorkspaceToolResponse struct {
-	Type      string          `jsonapi:"primary,tool"`
-	Success   bool            `jsonapi:"attr,success"`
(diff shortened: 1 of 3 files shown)


jrhouston (approved):
Thanks for fixing this @MulyaP

PR #467, opened 2026-08-12
add run tools integration tests and add guard for create_run tool

glbyktjys:
- Added integration tests for all run tools, comparing tool outputs with TFE API responses.
- Added a workspace lock check guard to the `create_run` tool.

### DIFF
--- a/pkg/tools/tfe/create_run.go
+++ b/pkg/tools/tfe/create_run.go
@@ -74,6 +74,10 @@ func createRunSafeHandler(ctx context.Context, request mcp.CallToolRequest, logg
 		return ToolErrorf(logger, "workspace '%s' not found in org '%s': %v", workspaceName, terraformOrgName, err)
 	}
 
+	if workspace.Locked {
+		return ToolErrorf(logger, "workspace '%s' is locked and cannot accept new runs. Use the force_unlock_workspace tool to unlock first", workspaceName)
+	}
+
 	options := &tfe.RunCreateOptions{
 		Workspace: workspace,
 	}
@@ -164,6 +168,10 @@ func createRunHandler(ctx context.Context, request mcp.CallToolRequest, logger *
 	workspace, err := tfeClient.Workspaces.Read(ctx, terraformOrgName, workspaceName)
 	if err != nil {
 		return ToolErrorf(logger, "workspace '%s' not found in org '%s': %v", workspaceName, terraformOrgName, err)
+	}
+
+	if workspace.Locked {
+		return ToolErrorf(logger, "workspace '%s' is locked and cannot accept new runs. Use the force_unlock_workspace tool to unlock first", workspaceName)
 	}
 
 	options := &tfe.RunCreateOptions{
--- a/pkg/tools/tfe/create_run.go
+++ b/pkg/tools/tfe/create_run.go
@@ -16,6 +16,8 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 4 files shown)


test/terraform/run_test.go
```diff
@@ -0,0 +1,304 @@
...
+}
+
+resource "terraform_data" "run_test" {
+  input = "terraform-mcp-server integration test for run tools"
+}
+
+output "run_test_id" {
+  value = terraform_data.run_test.id
+}
+`
```
jrhouston: Maybe we can create a `testdata` directory and store this as an actual `test.tf` file then use `go:embed` to embed the contents here. Storing terraform as strings inside the codebase is kinda gnarly.

test/terraform/run_test.go
```diff
@@ -0,0 +1,304 @@
...
+	})
+	require.NoError(t, err, "failed to create a configuration version")
+	require.NoError(t, client.ConfigurationVersions.UploadTarGzip(t.Context(), configurationVersion.UploadURL, &archive), "failed to upload the test configuration")
+}
+
+// waitForRun polls TFE until the run reaches the requested condition or times out.
+func waitForRun(t *testing.T, client *tfe.Client, runID, condition string, conditionMet func(*tfe.Run) bool) *tfe.Run {
+	t.Helper()
+
+	ctx, cancel := context.WithTimeout(t.Context(), 10*time.Minute)
```
jrhouston: 10 minutes seems like a long time? We can fail faster than that probably

pkg/tools/tfe/create_run.go
```diff
@@ -74,6 +74,10 @@ func createRunSafeHandler(ctx context.Context, request mcp.CallToolRequest, logg
 		return ToolErrorf(logger, "workspace '%s' not found in org '%s': %v", workspaceName, terraformOrgName, err)
 	}
 
+	if workspace.Locked {
+		return ToolErrorf(logger, "workspace '%s' is locked and cannot accept new runs. Use the force_unlock_workspace tool to unlock first", workspaceName)
```
jrhouston: `%q` here instead of `'%s'`

The error message is repeated verbatim twice. Maybe make it a const?

jrhouston (changes requested):
Couple small things

jrhouston:
Also you added tests for runs, you didn't actually add a test for the locking fix you introduced 😄

test/terraform/run_test.go
glbyktjys: yeah sounds good

test/terraform/run_test.go
```diff
@@ -0,0 +1,317 @@
...
+	require.NoError(t, err, "failed to lock test workspace")
+	defer client.Workspaces.ForceUnlock(t.Context(), workspace.ID)
+
+	result, resultText := callTool(t, s, "create_run", map[string]any{
+		"terraform_org_name": tfeOrgName,
+		"workspace_name":     workspaceName,
+	})
+
+	assert.True(t, result.IsError, "create_run should return an error for a locked workspace")
+	assert.Equal(t, `workspace "`+workspaceName+`" is locked and cannot accept new runs. Use the force_unlock_workspace tool to unlock first`, resultText)
```
jrhouston:
```suggestion
	assert.Equal(t, fmt.Sprintf(`workspace %q is locked and cannot accept new runs. Use the force_unlock_workspace tool to unlock first`, workspaceName), resultText)
```

jrhouston (changes requested):
New test looks good, one tiny nit.

jrhouston (approved):

PR #471, opened 2026-08-13
Add HCPT integration tests for stacks management tools

Marin2409:
### Changes

- `test/terraform/stacks_test.go` — `TestListStacks` and `TestGetStackDetails HCPT test integration
- `test/terraform/terraform_test.go` — adds `requireStacksEntitlement` helper that skips if the org does not have the Stacks permission; `Stacks.Create` returning `ErrResourceNotFound` also skips cleanly.
- `pkg/tools/tfe/list_stacks.go` — removes `ProjectName` from `StackSummary`; the `List` API only populates the project relation ID so the field was always returned empty.
- `pkg/tools/tfe/get_stack_details.go` — removes `terraform_org_name` as a required parameter; `Stacks.Read` only needs the `stack_id` and the org name was unused by the API call. Let me know if terraform_org_name had another function here, I can bring it back otherwise. 

## PCI review checklist


jrhouston (approved):
Excellent, thanks @Marin2409!

PR #472, opened 2026-08-13
add integration tests for private registry module and provider tools

glbyktjys:
## PCI review checklist


test/terraform/private_registry_test.go
```diff
@@ -0,0 +1,198 @@
...
+			return module
+		}
+
+		select {
+		case <-ctx.Done():
+			t.Fatalf("timed out waiting for private module %q version %s to finish processing", moduleLocator.Name, version)
+		case <-ticker.C:
+		}
+	}
+}
```
jrhouston: This is very similar logic to this other function for runs:

https://github.com/hashicorp/terraform-mcp-server/blob/45257192edfb390c1dc009ce2c566f1a920b9462/test/terraform/run_test.go#L288-L318

We could fashion a generic `waitFor` function that waits for things like:

```golang
func waitFor[T any](t *testing.T, duration time.Duration, description string, thingFunc func(context.Context) (*T, error)) *T
```

So we can do something like:
```golang
value := waitFor(t, 2*time.Minute, "wait for the registry module", func(ctx context.Context) (*tfe.TerraformRegistryModule, error) {
        module, err := client.RegistryModules.ReadTerraformRegistryModule(ctx, moduleLocator, version)
		if err == nil && module != nil && len(module.Root.Inputs) > 0 && len(module.Root.Outputs) > 0 && module.Root.Readme != "" {
			return module, nil
		}
        if err != nil {
            return nil, err
        }

        // keep waiting
        return nil, nil
})
```

We could also maybe use testify's eventually function which does something similar:

https://pkg.go.dev/github.com/stretchr/testify/assert#Eventually

test/terraform/private_registry_test.go
```diff
@@ -0,0 +1,198 @@
...
+			"private_module_id":      privateModuleAddress,
+			"private_module_version": moduleVersion,
+		})
+		require.False(t, result.IsError, "get_private_module_details should not return an error")
+		require.NotEmpty(t, resultText, "get_private_module_details response must not be empty")
+
+		// Verify against the TFE API directly.
+		expectedModuleAddress := strings.Join([]string{registryDetails.Namespace, registryDetails.Name, registryDetails.Provider}, "/")
+
+		// TODO: update this after MCP SKD migration
```
jrhouston, line 104: Yeah for real – this `.Contains` stuff is nasty.

jrhouston (changes requested):
Thanks for adding this @glbyktjys, just a note on the waiting logic.

glbyktjys:
- fixed hcpt test, `action_run` tool returned `str` instead of structured response so should use` assert.Contains()` instead of `gjson.Get()`.
- bumped up go and direct dependencies version for security scanning.

test/terraform/private_registry_test.go
glbyktjys: I slightly prefer `waitFor()` over `assert.Eventually()` because it provides clearer timeout errors. My only concern is that a generic helper adds another callback layer, such as waitForRun() wrapping waitFor(). A bit of complicated but that works.

jrhouston:
Needs rebase then it's good to go

jrhouston (approved):

PR #474, opened 2026-08-14
fix(action_run): update test assertions to match plain-text response

Marin2409:
The action_run handler was simplified to return a plain success message instead of a JSON object, because the JSON fields (success, run_id, message) were redundant they should return a simple success or failure message, not a full JSON object.

The test assertions that parsed "success" and "run_id" from the response are removed since those fields no longer exist. The post-apply check now uses the run ID that is already in scope rather than extracting it from the response.

## PCI review checklist

### DIFF
--- a/test/terraform/run_test.go
+++ b/test/terraform/run_test.go
@@ -204,7 +204,6 @@ func TestRunLifecycle(t *testing.T) {
 		assert.JSONEq(t, string(directJSON), resultText)
 	})
 
-	var actionRunID string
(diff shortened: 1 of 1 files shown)


glbyktjys:
Thank you! @Marin2409 I've fixed it here: https://github.com/hashicorp/terraform-mcp-server/pull/472/changes/227e1c169f0f5d2085a19fd965ac9ea04f15f690

jrhouston (approved):

PR #475, opened 2026-08-14
Add HCPT integration tests for `State Version` management tools

Marin2409:
### Changes

- **`test/terraform/state_version_test.go`** — adds four integration test functions:
  - `TestListStateVersionsHappyPath`
  - `TestListStateVersionsErrorPaths` 
  - `TestGetStateVersionHappyPath`
  - `TestGetStateVersionErrorPaths` 

- **`test/terraform/testdata/state_version_test.tf`** — dedicated test fixture with two `terraform_data` resources, used to assert state contents after apply

- **`test/terraform/terraform_test.go`** — moves `uploadConfiguration` here as a shared primitive; each test file provides its own thin wrapper (`uploadRunTestConfiguration`, `uploadStateVersionTestConfiguration`)

- **`pkg/tools/tfe/get_state_version.go`** — error messages now include the requested state version ID or workspace ID to aid debugging

## PCI review checklist

### DIFF
--- a/pkg/tools/tfe/get_state_version.go
+++ b/pkg/tools/tfe/get_state_version.go
@@ -42,8 +42,8 @@ func GetStateVersion(logger *log.Logger) server.ServerTool {
 
 // getStateVersionWithIDHandler handles tool logics and functionality
 func getStateVersionWithIDHandler(ctx context.Context, request mcp.CallToolRequest, logger *log.Logger) (*mcp.CallToolResult, error) {
-	stateVersionID := request.GetString("state_version_id", "")
-	workspaceID := request.GetString("workspace_id", "")
+	stateVersionID := GetTrimmedString(request, "state_version_id", "")
+	workspaceID := GetTrimmedString(request, "workspace_id", "")
 
 	stateVersionID = strings.TrimLeft(strings.TrimSpace(stateVersionID), "#")
 	workspaceID = strings.TrimLeft(strings.TrimSpace(workspaceID), "#")
@@ -57,7 +57,6 @@ func getStateVersionWithIDHandler(ctx context.Context, request mcp.CallToolReque
 	}
 
 	var sv *tfe.StateVersion
-
 	if stateVersionID == "" && workspaceID == "" {
 		return ToolError(logger, "One of state_version_id or workspace_id must be provided", nil)
 	}
@@ -75,6 +74,5 @@ func getStateVersionWithIDHandler(ctx context.Context, request mcp.CallToolReque
 	if err != nil {
 		return ToolError(logger, "failed to marshal state version", err)
 	}
-
 	return mcp.NewToolResultText(buf.String()), nil
 }
--- a/pkg/tools/tfe/get_state_version.go
+++ b/pkg/tools/tfe/get_state_version.go
@@ -32,12 +32,10 @@ func GetStateVersion(logger *log.Logger) server.ServerTool {
 				mcp.Description("Optional Workspace id to fetch latest version"),
 			),
 		),
-
 		Handler: func(ctx context.Context, request mcp.CallToolRequest) (*mcp.CallToolResult, error) {
 			return getStateVersionWithIDHandler(ctx, request, logger)
 		},
 	}
-
 }
 
 // getStateVersionWithIDHandler handles tool logics and functionality
--- a/pkg/tools/tfe/get_state_version.go
+++ b/pkg/tools/tfe/get_state_version.go
@@ -41,10 +41,12 @@ func GetStateVersion(logger *log.Logger) server.ServerTool {
 }
 
 // getStateVersionWithIDHandler handles tool logics and functionality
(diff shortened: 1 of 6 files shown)


pkg/tools/tfe/get_state_version.go
```diff
@@ -31,19 +32,19 @@ func GetStateVersion(logger *log.Logger) server.ServerTool {
...
-func getStateVersionWithIDHandler(
-	ctx context.Context,
-	request mcp.CallToolRequest,
-	logger *log.Logger) (*mcp.CallToolResult, error) {
+func getStateVersionWithIDHandler(ctx context.Context, request mcp.CallToolRequest, logger *log.Logger) (*mcp.CallToolResult, error) {
+	stateVersionID := GetTrimmedString(request, "state_version_id", "")
+	workspaceID := GetTrimmedString(request, "workspace_id", "")
+
+	stateVersionID = strings.TrimLeft(strings.TrimSpace(stateVersionID), "#")
+	workspaceID = strings.TrimLeft(strings.TrimSpace(workspaceID), "#")
```
jrhouston: I notice we are using GetTrimmedString above, but then also doing two more trim operations here. Can you explain why this is needed? Where does the `#` come from?

test/terraform/state_version_test.go
```diff
@@ -0,0 +1,222 @@
...
+}
+
+func TestListStateVersionsErrorPaths(t *testing.T) {
+	s := newTestingSession(t)
+	defer s.Close()
+
+	client := tfeClient(t)
+
+	nonExistentOrg := randomName("org-")
+	nonExistentWs := randomName("workspace-")
```
jrhouston: I don't think the non-existent ones need to be random every time. You can just hard code them as something like `org-doesnotexist123` like you did in the other tests.

test/terraform/state_version_test.go
```diff
@@ -0,0 +1,222 @@
...
+		Organization: tfeOrgName,
+		Workspace:    workspaceName,
+	})
+	require.NoError(t, err, "failed to list created state versions")
+	require.NotEmpty(t, stateVersions.Items, "applied run should create a state version")
+	stateVersion := stateVersions.Items[0]
+
+	t.Run("gets an exact state version by state_version_id", func(t *testing.T) {
+		result, resultText := callTool(t, s, "get_state_version", map[string]any{
+			"state_version_id": "#" + stateVersion.ID,
```
jrhouston: Why are we adding this `#` here if in the implementation we strip it away?

test/terraform/state_version_test.go
```diff
@@ -0,0 +1,222 @@
...
+		})
+
+		require.False(t, result.IsError, "get_state_version should not return an error")
+		require.NotEmpty(t, resultText, "get_state_version response must not be empty")
+		assert.Equal(t, stateVersion.ID, gjson.Get(resultText, "data.id").String())
+		assert.Equal(t, stateVersion.Serial, gjson.Get(resultText, "data.attributes.serial").Int())
+		assert.NotEmpty(t, gjson.Get(resultText, "data.attributes.created-at").String())
```
jrhouston, line 194: Perhaps we should have an assert that the resource from our test fixture is actually in the state returned?

https://github.com/hashicorp/terraform-mcp-server/blob/main/test/terraform/testdata/run_test.tf

For these tests I would consider duplicating that test fixture and putting a few test resources in it.

jrhouston (changes requested):
Couple questions

pkg/tools/tfe/get_state_version.go
Marin2409: The double TrimSpace must have happened when I added GetTrimmedString and then forgot to remove the existing strings.TrimSpace. That's my mistake.

I added the # trimming because, if the state version ID isn't already in context and the user retrieves it manually from the website, I wanted to guard against the case where they accidentally copy the sv-... ID with the hashtag included.

This is what updated it too:

```go
stateVersionID := GetTrimmedString(request, "state_version_id", "")
workspaceID := GetTrimmedString(request, "workspace_id", "")
```

test/terraform/state_version_test.go
Marin2409: Done, Updated non-existent values to hardcoded values

test/terraform/state_version_test.go
Marin2409: Good call, thanks for pointing that one out. Removed '#'

test/terraform/state_version_test.go
Marin2409: Added assert for resource testing.

pkg/tools/tfe/get_state_version.go
jrhouston: Yeah I think the `#` trimming is unnecessary because we've told the model the format in the tool definition, so it knows not to send it:
Screenshot 2026-08-17 at 6 08 07 PM

jrhouston (approved):

PR #476, opened 2026-08-14
add integration tests for `get_token_permissions`, `force_unlock_workspace`, and `get_sentinel_mock` tools

glbyktjys:
## PCI review checklist

### DIFF
--- a/pkg/tools/tfe/get_token_permissions.go
+++ b/pkg/tools/tfe/get_token_permissions.go
@@ -7,7 +7,6 @@ import (
 	"context"
 	"encoding/json"
 
-	"github.com/hashicorp/go-tfe"
 	"github.com/hashicorp/terraform-mcp-server/pkg/client"
 	"github.com/mark3labs/mcp-go/mcp"
 	"github.com/mark3labs/mcp-go/server"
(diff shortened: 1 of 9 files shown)


test/terraform/user_test.go
```diff
@@ -20,7 +21,65 @@ func TestWhoAmI(t *testing.T) {
...
+		Email            string `json:"email"`
+		IsServiceAccount *bool  `json:"is_service_account"`
+	}
+	require.NoError(t, json.Unmarshal([]byte(resultText), &toolAccount), "whoami should return valid account details")
+
+	tfeUser, err := client.Users.ReadCurrent(t.Context())
+	require.NoError(t, err, "failed to read the current user directly from TFE")
+	assert.Equal(t, tfeUser.Username, toolAccount.Username)
+	assert.Equal(t, tfeUser.Email, toolAccount.Email)
+	if assert.NotNil(t, toolAccount.IsServiceAccount) {
```
jrhouston: I don't think we need put `assert.NotNil` in a conditional here. `if toolAccount.IsServiceAccount != nil {}` is fine with the assertion we actually want to test below.

test/terraform/user_test.go
```diff
@@ -20,7 +21,65 @@ func TestWhoAmI(t *testing.T) {
...
+		"Traverse":                      tfeOrg.Permissions.CanTraverse,
+		"Update":                        tfeOrg.Permissions.CanUpdate,
+		"Update API Tokens":             tfeOrg.Permissions.CanUpdateAPIToken,
+		"Update OAuth":                  tfeOrg.Permissions.CanUpdateOAuth,
+		"Update Sentinel":               tfeOrg.Permissions.CanUpdateSentinel,
+		"Update HYOK Configuration":     tfeOrg.Permissions.CanUpdateHYOKConfiguration,
+		"View HYOK Feature Information": tfeOrg.Permissions.CanViewHYOKFeatureInfo,
+		"Enable Stacks":                 tfeOrg.Permissions.CanEnableStacks,
+		"Create Projects":               tfeOrg.Permissions.CanCreateProject,
+	}
```
jrhouston: This is almost a copypasta of the implementation: https://github.com/hashicorp/terraform-mcp-server/blob/407136375bcc64a40342ecf4ad6603eeab2f3e69/pkg/tools/tfe/get_token_permissions.go#L52-L70

jrhouston: Perhaps we should make this a const or export a function for mapping the permissions to human readable names – because this introduces two places where the list needs to be maintained.

jrhouston (changes requested):

test/terraform/user_test.go
glbyktjys: yeah you're right it's a bit redundant here.

test/terraform/user_test.go
glbyktjys: yeah since they are in different package, an export function should work.

pkg/tools/tfe/get_token_permissions.go
```diff
@@ -48,8 +49,21 @@ func getTokenPermissionsHandler(ctx context.Context, request mcp.CallToolRequest
...
+	buf, err := json.Marshal(HumanReadableTokenPermissions(org.Permissions))
+	if err != nil {
+		return ToolError(logger, "failed to marshal token permissions", err)
+	}
+
+	return mcp.NewToolResultText(string(buf)), nil
+}
+
+// HumanReadableTokenPermissions returns the display names of the permissions enabled for an organization.
+func HumanReadableTokenPermissions(permissions *tfe.OrganizationPermissions) []string {
```
jrhouston: Let's maybe move this function out of the `tools` package and into the `client` package, could put it here: https://github.com/hashicorp/terraform-mcp-server/blob/main/pkg/client/tfe_client.go

Having the tool export this function feels a bit weird.

pkg/tools/tfe/get_token_permissions_test.go
```diff
@@ -30,3 +31,20 @@ func TestGetTokenPermissions(t *testing.T) {
 		assert.Contains(t, tool.Tool.InputSchema.Required, "terraform_org_name")
 	})
 }
+
+func TestHumanReadableTokenPermissions(t *testing.T) {
```
jrhouston: Since this function is now the source of truth for what human readable names we translate these permissions to, the unit test case should cover all of 'em.

jrhouston (changes requested):
Couple more things then this is good to go 🚀

glbyktjys:
Increased the HCPT test timeout to [90 seconds](https://github.com/hashicorp/terraform-mcp-server/pull/476/changes#diff-5e167530ebd98a50ded0d697f08f90a8c140890e34c6c9f869cf32fcbd1a65caR21) and added polling for[state-version processing](https://github.com/hashicorp/terraform-mcp-server/pull/476/changes#diff-9dced46578454cef62a2c42669e965f1b7432a4d2158a76754b8e2961e1e4d80R62)to ensure state version metadata is processed before calling `list_state_versions`.

Because HCP Terraform processes uploads asynchronously, slower network or service processing can cause tests to run before resources are ready, resulting in flaky CI sometime. Added a slightly longer timeout to improve reliability. I can remove or adjust if they are unnecessary.

jrhouston (approved):

PR #480, opened 2026-08-17
Adding hcpt tests for variable set tools

MulyaP:
## PCI review checklist

### DIFF
--- a/pkg/tools/tfe/variable_sets.go
+++ b/pkg/tools/tfe/variable_sets.go
@@ -136,11 +136,14 @@ func CreateVariableSet(logger *log.Logger) server.ServerTool {
 				return ToolErrorf(logger, "failed to create variable set '%s' in org '%s': %v", name, orgName, err)
 			}
 
-			out, _ := json.Marshal(variableSetCreatedResult{
+			out, err := json.Marshal(variableSetCreatedResult{
 				VariableSetID:   varSet.ID,
 				VariableSetName: varSet.Name,
 				Message:         "variable set created successfully",
 			})
+			if err != nil {
+				return ToolError(logger, "failed to marshal create variable set response", err)
+			}
 			return &mcp.CallToolResult{
 				Content: []mcp.Content{
 					mcp.NewTextContent(string(out)),
@@ -203,12 +206,15 @@ func CreateVariableInVariableSet(logger *log.Logger) server.ServerTool {
 				return ToolErrorf(logger, "failed to create variable '%s' in variable set '%s': %v", key, varSetID, err)
 			}
 
-			out, _ := json.Marshal(variableCreatedResult{
+			out, err := json.Marshal(variableCreatedResult{
 				VariableID:    variable.ID,
 				VariableKey:   variable.Key,
 				VariableSetID: varSetID,
 				Message:       "variable created successfully",
 			})
+			if err != nil {
+				return ToolError(logger, "failed to marshal create variable response", err)
+			}
 			return &mcp.CallToolResult{
 				Content: []mcp.Content{
 					mcp.NewTextContent(string(out)),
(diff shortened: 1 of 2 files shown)


test/terraform/variable_sets_test.go
```diff
@@ -0,0 +1,270 @@
...
+	// what the tools under test do.
+	defer client.VariableSets.Delete(t.Context(), varSetID)
+
+	// Verify creation via the TFE API directly.
+	varSet, err := client.VariableSets.Read(t.Context(), varSetID, nil)
+	require.NoError(t, err, "variable set was reported as created but could not be read via the TFE API")
+	assert.Equal(t, varSetName, varSet.Name, "variable set name in TFE API should match the requested name")
+	assert.Equal(t, "Created by terraform-mcp-server integration tests", varSet.Description, "variable set description in TFE API should match")
+
+	// ── list_variable_sets ────────────────────────────────────────────────────
```
jrhouston: Comments like this are kinda unnecessary cus they just rephrase the `r.Run()` line below.

test/terraform/variable_sets_test.go
```diff
@@ -0,0 +1,270 @@
...
+	idx := strings.Index(s, prefix)
+	if idx == -1 {
+		return ""
+	}
+	rest := s[idx:]
+	if end := strings.IndexAny(rest, " \t\n\r"); end != -1 {
+		return rest[:end]
+	}
+	return rest
+}
```
jrhouston: Hmmmm why do we have a specialized function that parses free text and searches for an ID?

test/terraform/variable_sets_test.go
```diff
@@ -0,0 +1,270 @@
...
+		"name":               varSetName,
+		"description":        "Created by terraform-mcp-server integration tests",
+		"global":             false,
+	})
+	require.False(t, createResult.IsError, "create_variable_set should not return an error")
+	require.NotEmpty(t, createResultText, "create_variable_set result should not be empty")
+	assert.Contains(t, createResultText, varSetName, "create_variable_set response should reference the variable set name")
+
+	// Extract the variable set ID from the response text ("…with ID varset-xxxx").
+	// The response format is: "Successfully created variable set <name> with ID <id>"
```
jrhouston: Yeah I think in this case we should actually change the implementation here so it is structured to return the ID like `{"variable_set_id": "..."}`. 

The fact we are writing a special parsing function is kinda smelly, and making it structured will likely make it easier for the model to pull it out as well.

jrhouston (changes requested):

test/terraform/variable_sets_test.go
MulyaP: Agreed and removed em.

test/terraform/variable_sets_test.go
MulyaP: We can do that. I will go ahead and edit outputs for all tools which uses this natural language format to return ids, names, etc and a message like `Variable creation successful`

test/terraform/variable_sets_test.go
MulyaP: Removed that after changing the implementation to return json instead of natural language responses.

test/terraform/variable_sets_test.go
MulyaP: I changed the implementation for variable set tools, for other natural-language-returning tools, I aim to change them in a separate PR (if we want to). Resolving this for now.

pkg/tools/tfe/variable_sets.go
```diff
@@ -251,9 +267,14 @@ func AttachVariableSetToWorkspaces(logger *log.Logger) server.ServerTool {
 				return ToolErrorf(logger, "failed to attach variable set '%s' to workspaces: %v", varSetID, err)
 			}
 
+			out, _ := json.Marshal(map[string]interface{}{
+				"variable_set_id": varSetID,
+				"workspace_count": len(workspaces),
+				"message":         "variable set attached to workspaces successfully",
+			})
```
jrhouston: Let's make this a little `struct` with JSON tags like you did for the `whoami` tool

jrhouston (changes requested):
One last thing

pkg/tools/tfe/variable_sets.go
Mulya-Patel: Changed it. Sorry for that.

jrhouston (approved):

PR #481, opened 2026-08-18
Add HCPT integration tests for `Tags and Policy` management tools

Marin2409:
### Changes

- **`test/terraform/tags_and_policy_test.go`** — adds four integration test functions:
  - `TestListWorkspacePolicySets`
  - `TestListWorkspacePolicySetsErrorPaths` 
  - `TestAttachPolicySetToWorkspaces`
  - `TestAttachPolicySetToWorkspacesErrorPaths` 

- **`test/terraform/terraform_test.go`** — adds `requirePolicySetsEntitlement` helper, gated on
  `entitlements.Sentinel`; skips tests gracefully on orgs that do not have the policy sets feature enabled
  
## PCI review checklist

### DIFF
--- a/pkg/tools/tfe/policy_sets.go
+++ b/pkg/tools/tfe/policy_sets.go
@@ -17,6 +17,15 @@ import (
 	"github.com/mark3labs/mcp-go/server"
 )
 
+type MatchingPolicySet struct {
+	ID          string `json:"id"`
+	Name        string `json:"name"`
+	Description string `json:"description"`
+	Kind        string `json:"kind"`
+	Global      bool   `json:"global"`
+	Reason      string `json:"reason"`
+}
+
 // AttachPolicySetToWorkspaces creates a tool to attach a policy set to workspaces.
 func AttachPolicySetToWorkspaces(logger *log.Logger) server.ServerTool {
 	return server.ServerTool{
@@ -32,11 +41,11 @@ func AttachPolicySetToWorkspaces(logger *log.Logger) server.ServerTool {
 			),
 		),
 		Handler: func(ctx context.Context, request mcp.CallToolRequest) (*mcp.CallToolResult, error) {
-			policySetID, err := RequiredTrimmedString(request, "policy_set_id")
+			policySetID, err := RequireTrimmedString(request, "policy_set_id")
 			if err != nil {
 				return ToolError(logger, "missing required input: policy_set_id", err)
 			}
-			workspaceIDsStr, err := RequiredTrimmedString(request, "workspace_ids")
+			workspaceIDsStr, err := RequireTrimmedString(request, "workspace_ids")
 			if err != nil {
 				return ToolError(logger, "missing required input: workspace_ids", err)
 			}
@@ -97,11 +106,11 @@ func ListWorkspacePolicySets(logger *log.Logger) server.ServerTool {
 }
 
 func listWorkspacePolicySetsHandler(ctx context.Context, request mcp.CallToolRequest, logger *log.Logger) (*mcp.CallToolResult, error) {
-	terraformOrgName, err := RequiredTrimmedString(request, "terraform_org_name")
+	terraformOrgName, err := RequireTrimmedString(request, "terraform_org_name")
 	if err != nil {
 		return ToolError(logger, "missing required input: terraform_org_name", err)
 	}
-	workspaceID, err := RequiredTrimmedString(request, "workspace_id")
+	workspaceID, err := RequireTrimmedString(request, "workspace_id")
 	if err != nil {
 		return ToolError(logger, "missing required input: workspace_id", err)
 	}
@@ -109,6 +118,18 @@ func listWorkspacePolicySetsHandler(ctx context.Context, request mcp.CallToolReq
 	tfeClient, err := client.GetTfeClientFromContext(ctx, logger)
 	if err != nil {
 		return ToolError(logger, "failed to get Terraform client - ensure TFE_TOKEN and TFE_ADDRESS are configured", err)
+	}
+
+	workspace, err := tfeClient.Workspaces.ReadByID(ctx, workspaceID)
+	if err != nil {
+		return ToolErrorf(logger, "workspace not found: %s", workspaceID)
+	}
+
+	// A workspace belongs to exactly one org and policy sets are org-scoped, so a
+	// mismatch would report this org's global policy sets against a foreign workspace.
+	if workspace.Organization != nil && !strings.EqualFold(workspace.Organization.Name, terraformOrgName) {
+		return ToolErrorf(logger, "workspace %q belongs to organization %q, not %q",
+			workspaceID, workspace.Organization.Name, terraformOrgName)
 	}
 
 	// Paginate through all policy sets with the workspaces included
@@ -181,13 +202,4 @@ func listWorkspacePolicySetsHandler(ctx context.Context, request mcp.CallToolReq
 			mcp.NewTextContent(string(result)),
 		},
 	}, nil
-}
-
-type MatchingPolicySet struct {
-	ID          string `json:"id"`
-	Name        string `json:"name"`
-	Description string `json:"description"`
-	Kind        string `json:"kind"`
-	Global      bool   `json:"global"`
-	Reason      string `json:"reason"`
 }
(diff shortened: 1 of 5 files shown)


pkg/tools/tfe/policy_sets.go
```diff
@@ -104,25 +106,28 @@ func listWorkspacePolicySetsHandler(ctx context.Context, request mcp.CallToolReq
 		return ToolError(logger, "missing required input: workspace_id", err)
 	}
 
+	terraformOrgName = strings.TrimSpace(terraformOrgName)
```
jrhouston: For these you could make an analog to your `GetTrimmedString()` helper called `RequireTrimmedString()`

https://github.com/hashicorp/terraform-mcp-server/blob/846e172dcbcbe33b8aadfb4cfd5fa5098322e114/pkg/tools/tfe/strings.go#L12-L16

pkg/tools/tfe/policy_sets.go
```diff
@@ -104,25 +106,28 @@ func listWorkspacePolicySetsHandler(ctx context.Context, request mcp.CallToolReq
...
+		policySets, err := tfeClient.PolicySets.List(ctx, terraformOrgName, &tfe.PolicySetListOptions{
 			Include: []tfe.PolicySetIncludeOpt{tfe.PolicySetWorkspaces},
 			ListOptions: tfe.ListOptions{
 				PageNumber: pageNumber,
 				PageSize:   100,
 			},
 		})
 		if err != nil {
-			return ToolErrorf(logger, "failed to list policy sets for org '%s': %v", orgName, err)
+			return ToolErrorf(logger, "failed to list policy sets for org '%s': %v", terraformOrgName, err)
```
jrhouston: Since you're changing this use `%q` here.

test/terraform/tags_and_policy_test.go
```diff
@@ -0,0 +1,208 @@
...
+	ps, err := client.PolicySets.Create(t.Context(), tfeOrgName, tfe.PolicySetCreateOptions{
+		Name: tfe.String(psName),
+	})
+	require.NoError(t, err, "setup: failed to create policy set via TFE API")
+	defer client.PolicySets.Delete(t.Context(), ps.ID)
+
+	t.Run("attaches policy set to multiple workspaces", func(t *testing.T) {
+		result, resultText := callTool(t, s, "attach_policy_set_to_workspaces", map[string]any{
+			"policy_set_id": ps.ID,
+			"workspace_ids": " " + workspaceIDs[0] + ", " + workspaceIDs[1] + " ",
```
jrhouston:
```suggestion
			"workspace_ids": strings.Join(workspaceIDs, ", "),
```

jrhouston (changes requested):
Couple small things

pkg/tools/tfe/policy_sets.go
Marin2409: Helper method `RequireTrimmedString` added 👍🏾

pkg/tools/tfe/policy_sets.go
Marin2409: Updated to %q

test/terraform/tags_and_policy_test.go
Marin2409: Updated to `strings.Join()`

test/terraform/tags_and_policy_test.go
```diff
@@ -0,0 +1,209 @@
...
+		// The tool never validates whether workspace_id exists in the API — it uses it
+		// only as a client-side filter against directly-attached policy sets. Global
+		// policy sets are always returned regardless. The response content therefore
+		// depends on live org state and cannot be deterministically asserted here.
+		result, resultText := callTool(t, s, "list_workspace_policy_sets", map[string]any{
+			"terraform_org_name": tfeOrgName,
+			"workspace_id":       nonExistentWsID,
+		})
+		assert.False(t, result.IsError, "list_workspace_policy_sets should not return an error for a non-existent workspace ID")
+		assert.Contains(t, resultText, "No policy sets are attached to workspace", "response should indicate no policy sets are attached")
```
jrhouston: I found this test and the comment above confusing. 

We're just saying that a quirk of the API is that it will accept a workspaceID that doesn't exist, and instead of returning an error it will just return that there's no attachments to that workspaceID?

If so this is a strange thing to codify in our tests. We could just add the logic to the tool to explicitly check if the workspace exists, then return an error. That seems like a reasonable thing to do.

test/terraform/tags_and_policy_test.go
```diff
@@ -0,0 +1,209 @@
...
+		globalPsResult := gjson.Get(resultText, fmt.Sprintf("#(id==%q)", globalPs.ID))
+		require.True(t, globalPsResult.Exists(), "response should contain the global policy set")
+		assert.Equal(t, globalPsName, globalPsResult.Get("name").String(), "global policy set name should match")
+		assert.Equal(t, "global", globalPsResult.Get("reason").String(), "global policy set should have reason 'global'")
+		assert.True(t, globalPsResult.Get("global").Bool(), "global policy set should have global=true")
+	})
+}
+
+func TestListWorkspacePolicySetsErrorPaths(t *testing.T) {
+	s := newTestingSession(t)
```
jrhouston, line 96: I think you need the `requirePolicySetsEntitlement(t)` call in this test too?

pkg/tools/tfe/policy_sets.go
```diff
@@ -175,10 +176,18 @@ func listWorkspacePolicySetsHandler(ctx context.Context, request mcp.CallToolReq
...
 	}
-
 	return &mcp.CallToolResult{
 		Content: []mcp.Content{
 			mcp.NewTextContent(string(result)),
 		},
 	}, nil
 }
+
+type MatchingPolicySet struct {
```
jrhouston: I know you've moved this to be consistent with other files but the convention in Go is actually to have them at the top of the file – I would just remove this change from the PR tbh.

pkg/tools/tfe/strings.go
```diff
@@ -14,3 +14,12 @@ func GetTrimmedString(request mcp.CallToolRequest, name, defaultValue string) st
 	value := request.GetString(name, defaultValue)
 	return strings.TrimSpace(value)
 }
+
+// RequiredTrimmedString wrapper: function around request.RequireString that also calls TrimSpace()
+func RequiredTrimmedString(request mcp.CallToolRequest, name string) (string, error) {
```
jrhouston: nit: this function wraps `RequireString`, and `GetTrimmedString` wraps `GetString`, so calling it`RequireTrimmedString` would be more consistent.

jrhouston (changes requested):
Spotted a couple more things

pkg/tools/tfe/strings.go
Marin2409: Renamed to `RequireTrimmedString` 👍🏾

pkg/tools/tfe/policy_sets.go
Marin2409: Moved back to its original place

test/terraform/tags_and_policy_test.go
Marin2409: Not on that one, the entitlement guard belongs only where the test creates resources that require the feature, for example,  `TestListWorkspacePolicySets` and `TestAttachPolicySetToWorkspaces` where `client.PolicySets.Create(...)` would fail without it.

`TestListWorkspacePolicySetsErrorPaths` uses a fake org name so the `PolicySets.List` API call fails immediately  before the policy sets feature is ever checked.

test/terraform/tags_and_policy_test.go
Marin2409: You're right, I agree. The tool was only using workspace_id as a client-side filter against directly-attached policy sets, so an unknown ID fell through and returned just the org's global policy sets. 

So I've added the explicit check you suggested:

```go
workspace, err := tfeClient.Workspaces.ReadByID(ctx, workspaceID)
if err != nil {
    return ToolErrorf(logger, "workspace not found: %s", workspaceID)
}
```

One related thing I noticed while wiring it up: ReadByID isn't org-scoped, so a real workspace ID from a different org still passed the check and came back with this org's global policy sets attributed to it. The org name is already in that same response, so the guard is free:

```go
if workspace.Organization != nil && !strings.EqualFold(workspace.Organization.Name, terraformOrgName) {
    return ToolErrorf(logger, "workspace %q belongs to organization %q, not %q",
        workspaceID, workspace.Organization.Name, terraformOrgName)
}
```

The test now asserts the error and its message instead of the old no-op behavior, plus a new subtest for the cross-org case.

jrhouston:
@Marin2409 this is good to go, can you rebase it?

Marin2409:
> @Marin2409 this is good to go, can you rebase it?

Rebased 👍🏾

jrhouston (approved):

PR #487, opened 2026-08-21
Add stale issue and PR pruning

jrhouston:
Adding stale issue and PR pruning to the project as suggested by @glbyktjys

Also, I've updated the changelog workflow so it doesn't pester us unless we actually modify the application code.

## PCI review checklist

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -9,6 +9,7 @@ on:
   workflow_dispatch:
 
 permissions:
+  actions: write
   contents: read
   issues: write
   pull-requests: write
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,39 @@
+# Copyright (c) HashiCorp, Inc.
+# SPDX-License-Identifier: MPL-2.0
+
+name: Prune stale issues and PRs
+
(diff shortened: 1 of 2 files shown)


.github/workflows/stale.yml
```diff
@@ -0,0 +1,39 @@
...
+# SPDX-License-Identifier: MPL-2.0
+
+name: Prune stale issues and PRs
+
+on:
+  schedule:
+    - cron: "0 0 * * *"
+  workflow_dispatch:
+
+permissions:
```
glbyktjys, line 11: I see in v11 actions/stale it has recommended permissions of [add `action: write`](https://github.com/marketplace/actions/close-stale-issues#close-stale-issues-and-prs)
not entirely sure if we need to add this here

glbyktjys (dismissed):
Thank you for adding this! The rest looks great to me

.github/workflows/stale.yml
jrhouston: It's because the action [stores some state and writes](https://github.com/marketplace/actions/close-stale-issues#statefulness) to the Actions API to do that. Have updated to add the permission.

glbyktjys (approved):

PR #493, opened 2026-08-24
[Official SDK Migration] Refactor tools and client into packages and add `list_terraform_orgs` tool

MulyaP:
This PR adds `list_terraform_org` tool using the official go-sdk for MCP and refactors the official sdk code into packages like in mark2labs/mcp-go code.

## PCI review checklist

### DIFF
--- a/pkg/mcp-official/tools/tfe/organizations.go
+++ b/pkg/mcp-official/tools/tfe/organizations.go
@@ -7,7 +7,6 @@ import (
 
 	"github.com/hashicorp/go-tfe"
 	"github.com/hashicorp/terraform-mcp-server/pkg/mcp-official/client"
-	"github.com/hashicorp/terraform-mcp-server/pkg/mcp-official/utils"
 	"github.com/modelcontextprotocol/go-sdk/mcp"
 )
 
@@ -38,7 +37,7 @@ func ListTerraformOrganizationsTool() *mcp.Tool {
 		Annotations: &mcp.ToolAnnotations{
 			Title:           "List all Terraform organizations",
 			ReadOnlyHint:    true,
-			DestructiveHint: utils.Ptr(false),
+			DestructiveHint: ptr(false),
 		},
 	}
 }
@@ -76,4 +75,8 @@ func ListTerraformOrganizationsFunc(ctx context.Context, request *mcp.CallToolRe
 		Items:      summaries,
 		Pagination: orgs.Pagination,
 	}, nil
+}
+
+func ptr[T any](v T) *T {
+    return &v
 }
--- a/pkg/mcp-official/tools/tfe/organizations.go
+++ b/pkg/mcp-official/tools/tfe/organizations.go
@@ -1,4 +1,4 @@
-package mcpofficial
+package tools
 
 import (
 	"context"
@@ -8,6 +8,7 @@ import (
 	"github.com/hashicorp/go-tfe"
 	"github.com/modelcontextprotocol/go-sdk/mcp"
 	log "github.com/sirupsen/logrus"
+	"github.com/hashicorp/terraform-mcp-server/pkg/mcp-official/client"
 )
 
 type OrganizationSummary struct {
@@ -45,7 +46,7 @@ func ListTerraformOrganizationsTool() *mcp.Tool {
 func ListTerraformOrganizationsFunc(ctx context.Context, request *mcp.CallToolRequest, input ListOrganizationsArguments) (*mcp.CallToolResult, *OrganizationSummaryList, error) {
 	log.Info("ListTerraformOrganizations for official mcp go-sdk called..")
 
-	tfeClient, err := GetTfeClient(ctx)
+	tfeClient, err := client.GetTfeClient(ctx)
 	if err != nil {
 		return nil, nil, err
 	}
--- a/pkg/mcp-official/tools/tfe/organizations.go
+++ b/pkg/mcp-official/tools/tfe/organizations.go
@@ -6,11 +6,12 @@ import (
 	"time"
 
 	"github.com/hashicorp/go-tfe"
-	"github.com/modelcontextprotocol/go-sdk/mcp"
-	log "github.com/sirupsen/logrus"
 	"github.com/hashicorp/terraform-mcp-server/pkg/mcp-official/client"
+	"github.com/hashicorp/terraform-mcp-server/pkg/mcp-official/utils"
+	"github.com/modelcontextprotocol/go-sdk/mcp"
 )
 
+// OrganizationSummary holds a trimmed view of a single Terraform organization.
 type OrganizationSummary struct {
 	Name      string    `json:"organization_name"`
 	Email     string    `json:"organization_email"`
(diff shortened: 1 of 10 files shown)


pkg/mcp-official/tools/tfe/organizations.go
```diff
@@ -0,0 +1,80 @@
...
+
+type ListOrganizationsArguments struct {
+	// Optional fields (will be zero values if not provided)
+	Page     int `json:"page,omitempty" jsonschema:"Page number for pagination (min 1)"`
+	PageSize int `json:"pageSize,omitempty" jsonschema:"Results per page for pagination (min 1, max 100)"`
+}
+
+func ListTerraformOrganizationsTool() *mcp.Tool {
+	trueVal := true
+	falseVal := false
```
jrhouston: This is gross and no doubt we're gonna need to do it in a bunch of places. Instead, we could make a helper func for this like:

```golang
func ptr[T any](v T) *T {
    return &v
}
```

pkg/mcp-official/tools/tfe/organizations.go
```diff
@@ -0,0 +1,80 @@
...
+	"fmt"
+	"time"
+
+	"github.com/hashicorp/go-tfe"
+	"github.com/modelcontextprotocol/go-sdk/mcp"
+	log "github.com/sirupsen/logrus"
+	"github.com/hashicorp/terraform-mcp-server/pkg/mcp-official/client"
+)
+
+type OrganizationSummary struct {
```
jrhouston, line 14: With the new code let's be consistent and put the godoc comments above every struct definition

pkg/mcp-official/tools/tfe/organizations.go
```diff
@@ -0,0 +1,80 @@
...
+	*tfe.Pagination
+}
+
+type ListOrganizationsArguments struct {
+	// Optional fields (will be zero values if not provided)
+	Page     int `json:"page,omitempty" jsonschema:"Page number for pagination (min 1)"`
+	PageSize int `json:"pageSize,omitempty" jsonschema:"Results per page for pagination (min 1, max 100)"`
+}
+
+func ListTerraformOrganizationsTool() *mcp.Tool {
```
jrhouston, line 33: Something that's missing here is injecting the logger dependency. 

It gets passed in at the `RegisterTools()` call here: https://github.com/hashicorp/terraform-mcp-server/blob/main/pkg/mcp-official/tools.go#L9 

But then it just gets swallowed, and the code here defaults to using `log` from the import.

pkg/mcp-official/tools/tfe/workspace.go
```diff
@@ -53,7 +54,7 @@ func ListWorkpsacesTool() *mcp.Tool {
 }
 
 func ListWorkspacesFunc(ctx context.Context, request *mcp.CallToolRequest, input ListWorkspacesArguments) (*mcp.CallToolResult, *WorkspaceSummaryList, error) {
-	log.Info("ListWorkspaces for official mcp go-dk called..")
+	log.Info("ListWorkspaces for official mcp go-sdk called..")
```
jrhouston: This log message isn't very useful, there is already a tool call logger middleware that's going to get ported across: https://github.com/Marin2409/terraform-mcp-server/blob/main/pkg/client/tool_logging_middleware.go

jrhouston (changes requested):

pkg/mcp-official/tools/tfe/organizations.go
MulyaP: I looked into this and simply passing logger into the handler func argument in `mcp.AddTool` will not work because of how of its designed. One solution that works here is to make the handler a closure function which looks something like this:

```
func ListTerraformOrganizationsFunc(logger *log.Logger) func(ctx context.Context, req *mcp.CallToolRequest, input ListOrganizationsArguments) (*mcp.CallToolResult, *OrganizationSummaryList, error) {
	return func(ctx context.Context, req *mcp.CallToolRequest, input ListOrganizationsArguments) (*mcp.CallToolResult, *OrganizationSummaryList, error) {
		log.Info("ListTerraformOrganizations for official mcp go-sdk called.. dkfhgbdf")
	}
}
```

In my opinion, this makes it unnecessarily convoluted and importing log seems like a good alternative. I will try to find other methods with which we can achieve this in a simpler way.

pkg/mcp-official/tools/tfe/workspace.go
MulyaP: I agree, I should remove it. It was just there for me to know which tool endpoint was called (official or mark3labs) when testing it.

pkg/mcp-official/tools/tfe/organizations.go
MulyaP: Agreed and added

pkg/mcp-official/tools/tfe/organizations.go
MulyaP: Agree. Made a utils folder and utils.go file and added the helper function there.

jrhouston: So having a `utils` package is kinda a smell in Go, see the Google Go Style Guide:

> Avoid uninformative package names like util, utility, common, helper, model, testhelper, and so on that would tempt users of the package to [rename it when importing](https://google.github.io/styleguide/go/decisions.html#import-renaming). See: https://google.github.io/styleguide/go/decisions.html#package-names

I'd just keep this func in the package it's used in for now.

pkg/mcp-official/tools/tfe/organizations.go
jrhouston: Yeah I agree it's annoying – but if we only import the log package then we lose the logger we configure when the server gets initialized, and the logger has a bunch of options around formatting and log levels: https://github.com/hashicorp/terraform-mcp-server/blob/main/cmd/terraform-mcp-server/init.go#L224-L249

pkg/mcp-official/tools/tfe/organizations.go
Mulya-Patel: Okay! Sorry, didn't know this. Will just remove utils stuff and add that function to tools.go.

pkg/mcp-official/tools/tfe/organizations.go
Mulya-Patel: Oh, I wasn't aware of this initialization. That kind of leaves us no choice but to go through the closure function method. Will make this change to both the tools.

jrhouston: A possible alternative is to create a [tool middleware](https://pkg.go.dev/github.com/modelcontextprotocol/go-sdk/mcp#Middleware) which uses [ctx.WithValue()](https://pkg.go.dev/context#WithValue) to attach the logger to the context of each tool call. Then we can use [ctx.Value()](https://pkg.go.dev/context#example-WithValue) to grab the logger when we need it.

jrhouston:
@Mulya-Patel can you rebase this?

jrhouston (approved):

Issue #438, opened 2026-07-31
retrieve private registry submodule information


cmylchan:
currently the `get_private_module_details` tool only returns top-level module info and does not expose any additional information at the individual submodule level

expected outcome: expose private registry submodule information in the same manner as top-level modules.

PR #495, opened 2026-08-25
feat(tfe): expose private registry submodule details in `get_private_module_details`

Marin2409:
This PR closes issue [#438](https://github.com/hashicorp/terraform-mcp-server/issues/438) where `get_private_module_details` only returned top-level module information. Private registry submodules are now reported the same way as the root module, each with their own inputs, outputs, dependencies, provider dependencies, resources, and README.

## Changes

- Decoded the registry response into `client.TerraformModuleVersionDetails` instead of `tfe.TerraformRegistryModule`, which has no `Submodules` field and was silently discarding the data.
- Reported each submodule through the same formatting helpers as the root module.
- Labelled the root section `Root Module:` and moved its README there, instead of leaving it at the bottom where nothing identified which module it belonged to.
- Added a `Dependencies` table, applying to the root module and all submodules.
- Reported the module version, and pinned the usage example to the version the details came from rather than the first `version-statuses` entry, which may be unpublished or failed.
- Requested the documented latest-version route when `private_module_version` is omitted, rather than sending an empty version segment.
- Updated the tool description to mention submodules.
- Extended the integration test to cover explicit and omitted versions, added a `modules/child` fixture, and made polling failures report which part never finished processing.

## Workflow

1. Use `search_private_modules` to find the exact `private_module_id`.
2. Call `get_private_module_details`, optionally passing `private_module_version`. When omitted, the registry resolves the latest version and the response states which one it used.
3. The response contains the usage example, basic information, a `Root Module:` section, and one `Submodule:` section per nested module.

## PCI review checklist

### DIFF
--- a/pkg/tools/tfe/get_private_module_details.go
+++ b/pkg/tools/tfe/get_private_module_details.go
@@ -6,6 +6,7 @@ package tools
 import (
 	"context"
 	"fmt"
+	"net/url"
 	"path"
 	"regexp"
 	"strings"
@@ -94,14 +95,14 @@ func getPrivateModuleDetailsHandler(ctx context.Context, request mcp.CallToolReq
 	}).Info("Getting private module details")
 
 	var module *tfe.RegistryModule
-	var terraformRegistryModule *tfe.TerraformRegistryModule
+	var terraformRegistryModule *client.TerraformModuleVersionDetails
 
 	module, err = tfeClient.RegistryModules.Read(ctx, tfeModuleID)
 	if err != nil {
 		return ToolErrorf(logger, "module not found: %s - use search_private_modules to find valid module IDs", moduleID)
 	}
 
-	terraformRegistryModule, err = tfeClient.RegistryModules.ReadTerraformRegistryModule(ctx, tfeModuleID, moduleVersion)
+	terraformRegistryModule, err = readTerraformRegistryModuleDetails(ctx, tfeClient, tfeModuleID, moduleVersion)
 	if err != nil {
 		logger.WithError(err).Warn("failed to get detailed module information from Terraform Registry, continuing with basic info")
 	}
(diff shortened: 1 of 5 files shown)


pkg/tools/tfe/get_private_module_details.go
```diff
@@ -22,9 +23,7 @@ import (
...
-It provides comprehensive details including inputs, outputs, dependencies, versions, and usage examples. The private_module_id format is 'module-namespace/module-name/module-provider-name'.
-This can be obtained by calling 'search_private_modules' first to obtain the exact private_module_id required to use this tool. This tool requires a valid Terraform token to be configured.`),
+			mcp.WithDescription(`This tool retrieves detailed information about a specific private module in your Terraform Cloud/Enterprise organization. It provides comprehensive details including inputs, outputs, dependencies, versions, and usage examples. It also reports every submodule (nested module) published with the module, each with its own inputs, outputs, provider dependencies, resources, and README. The private_module_id format is 'module-namespace/module-name/module-provider-name'. This can be obtained by calling 'search_private_modules' first to obtain the exact private_module_id required to use this tool. This tool requires a valid Terraform token to be configured.`),
```
MulyaP: I feel like we don't need the last line (`This tool requires a valid Terraform token to be configured.`) since the tool won't be registered if terraform token is not present.

pkg/tools/tfe/get_private_module_details.go
```diff
@@ -94,40 +89,73 @@ func getPrivateModuleDetailsHandler(ctx context.Context, request mcp.CallToolReq
...
+	if moduleID.RegistryName == tfe.PublicRegistry {
+		basePath = "/api/registry/public/v1/modules"
+	}
+
+	u := fmt.Sprintf("%s/%s/%s/%s",
+		basePath,
+		url.PathEscape(moduleID.Namespace),
+		url.PathEscape(moduleID.Name),
+		url.PathEscape(moduleID.Provider),
+	)
```
jrhouston, line 118: Nit: you could use `path.Join()` here instead.

jrhouston (approved):

PR #497, opened 2026-08-26
Migrate official go-sdk server setup to prepare for middleware and hooks

Shweta-Deshpande:
This PR cleans up how we build the official MCP go-sdk server (pkg/mcp-official), as a first step before we add middleware and session hooks in follow-up PRs.

**What changed:**
- NewServer now takes version and instructions directly as arguments instead of importing the version package itself. 
- Added a small internal serverConfig struct plus an Option pattern (like mcp.NewServer(..., opts...)). This doesn't change any behavior today — it just gives us a clean place to plug in middleware and session hooks later.
- Removed an unused tlsConfig parameter from getOfficialStreamableServer.
- Hooked up a logger to the go-sdk's StreamableHTTPOptions, so the SDK's internal logs go through via newSlogLogger.

**We're planning two follow-up changes:**
1. Middleware — adding request/response middleware (e.g. logging, auth checks) via AddReceivingMiddleware.
2. Hooks — running custom logic when a session starts/ends (onSession hooks).

## PCI review checklist

### DIFF
--- a/cmd/terraform-mcp-server/init.go
+++ b/cmd/terraform-mcp-server/init.go
@@ -368,8 +368,8 @@ func streamableHTTPServerInit(ctx context.Context, hcServer *server.MCPServer, l
 
 	// Create the official go-sdk streamable server
 	if enableOfficialSDK := os.Getenv("TF_X_OFFICIAL_SDK_ENABLED"); enableOfficialSDK == "true" {
-		logger.Infof("TF_X_OFFICIAL_SDK_ENABLED set to true in env, enabling the official mcp go-sdk server")
-		officialStreamableServer := getOfficialStreamableServer(ctx, heartbeatInterval, isStateless, tlsConfig, corsConfig, logger, organizationAllowlist, enabledToolsets)
+		logger.Info("TF_X_OFFICIAL_SDK_ENABLED set to true in env, enabling the official mcp go-sdk server")
+		officialStreamableServer := getOfficialStreamableServer(ctx, heartbeatInterval, isStateless, corsConfig, logger, organizationAllowlist, enabledToolsets)
 		// Handle the /mcp endpoint with the official go-sdk streamable server (with security wrapper)
 		mux.Handle(endpointPath+"/official", officialStreamableServer)
 		mux.Handle(endpointPath+"/official/", officialStreamableServer)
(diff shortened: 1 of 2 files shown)


cmd/terraform-mcp-server/init.go
```diff
@@ -458,13 +458,14 @@ func streamableHTTPServerInit(ctx context.Context, hcServer *server.MCPServer, l
 	return nil
 }
 
-func getOfficialStreamableServer(ctx context.Context, heartbeatInterval time.Duration, isStateless bool, tlsConfig *client.TLSConfig, corsConfig client.CORSConfig, logger *log.Logger, organizationAllowlist []string, enabledToolsets []string) http.Handler {
-	logger.Infof("Creating a go-sdk StreamableHTTP server...")
-	hcServer := mcpofficial.NewServer(heartbeatInterval, logger, enabledToolsets)
+func getOfficialStreamableServer(ctx context.Context, heartbeatInterval time.Duration, isStateless bool, corsConfig client.CORSConfig, logger *log.Logger, organizationAllowlist []string, enabledToolsets []string) http.Handler {
```
jaylonmcshan19-x, line 461: `organizationAllowlist` gets passed into `getOfficialStreamableServer` but NewServer doesn't
take it, so nothing on that path uses it right now. guessing that's the phase 2
middleware work correct? I'm asking due to all six team tools I'm migrating for `13099` take
`terraform_org_name`, so it'll be helpful to know whether to check it in the tool funcs for now or
just leave it til the middleware lands.

jaylonmcshan19-x (approved):

cmd/terraform-mcp-server/init.go
Shweta-Deshpande: > `organizationAllowlist` gets passed into `getOfficialStreamableServer` but NewServer doesn't take it, so nothing on that path uses it right now. guessing that's the phase 2 middleware work correct? I'm asking due to all six team tools I'm migrating for `13099` take `terraform_org_name`, so it'll be helpful to know whether to check it in the tool funcs for now or just leave it til the middleware lands.

Thanks @jaylonmcshan19-x you are right, the middleware work will cover the org allowed list checks. All those tools just take in the terraform_org_name and call the tfe api. The check for allowed orgs is centralized in the OrganizationAllowlistToolMiddleware.

PR #500, opened 2026-08-27
Migrate server.ClientSession mark3labs type to session ID string

Shweta-Deshpande:
## What changed

Change the four session-scoped helper functions in `pkg/client` (`NewSessionHandler`, `EndSessionHandler`, `CreateTfeClientForSession`, `CreateHttpClientForSession`) to take a plain `sessionID string` instead of the mark3labs `server.ClientSession` interface. Also updated the calls in `cmd/terraform-mcp-server/main.go` accordingly, and dropped the`server` import from `pkg/client/session.go`.

## Why

We're migrating the server setup, middleware, and session hooks to the official MCP go-sdk (see #497 for the first step). The go-sdk represents a session as `*mcp.ServerSession`, a different type from mark3labs' `server.ClientSession` — and neither implements the other.

Since these 4 functions never actually needed anything from `server.ClientSession` besides the ID, keeping that type in their signature would force the upcoming middleware and session-hooks PRs to build a fake `server.ClientSession` just to call them. Narrowing the parameter to a `string` removes that dependency, so both the old mark3labs code path and the new go-sdk path can call the same functions directly with their own session's ID.

Doing this as its own small PR keeps it easy to review in isolation and easy to revert independently of the middleware/hooks work that follows.

## PCI review checklist

### DIFF
--- a/cmd/terraform-mcp-server/main.go
+++ b/cmd/terraform-mcp-server/main.go
@@ -43,7 +43,7 @@ func runHTTPServer(logger *log.Logger, host string, port string, endpointPath st
 	// Create hooks for session management
 	hooks := &server.Hooks{}
 	hooks.AddOnRegisterSession(func(ctx context.Context, session server.ClientSession) {
-		client.NewSessionHandler(ctx, session, logger)
+		client.NewSessionHandler(ctx, session.SessionID(), logger)
 	})
 
 	serverOpts := []server.ServerOption{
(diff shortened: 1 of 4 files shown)


jaylonmcshan19-x (approved):

jaylonmcshan19-x:
@Shweta-Deshpande narrowing to the ID looks good to me! separate thing I noticed. mcp-official's `GetTfeClient` wraps everything in a `sync.Once`, so whoever hits that endpoint first creates the client and everyone after that reuses it. it pulls the token off the context, so in `streamable-http` that looks like the same token bleed you fixed in [#395](https://github.com/hashicorp/terraform-mcp-server/pull/395). behind the feature flag so nothing's exposed right now, but every tool we migrate ends up calling it. is that already on the list for the migration or should I track it?

Shweta-Deshpande:
> @Shweta-Deshpande narrowing to the ID looks good to me! separate thing I noticed. mcp-official's `GetTfeClient` wraps everything in a `sync.Once`, so whoever hits that endpoint first creates the client and everyone after that reuses it. it pulls the token off the context, so in `streamable-http` that looks like the same token bleed you fixed in [#395](https://github.com/hashicorp/terraform-mcp-server/pull/395). behind the feature flag so nothing's exposed right now, but every tool we migrate ends up calling it. is that already on the list for the migration or should I track it?

Good catch @jaylonmcshan19-x. I have this on my radar for the migration eventually. Will bring this change in, in the upcoming PRs.

PR #501, opened 2026-08-27
[Official SDK Migration] Migrate server middleware: rate limiting, tool-call logging, and org allowlist

Shweta-Deshpande:
## What changed

Adds the go-sdk versions of the three tool-call checks  - rate limiting, tool-call logging, organization allowlist

- pkg/mcp-official: three new middlewares — RateLimitMiddleware, ToolLoggingMiddleware, OrganizationAllowlistMiddleware — plus a WithMiddlewares option so NewServer can attach them.

- pkg/client: Added reuse of logic for both packages:

1. RateLimitMiddleware.Allow(sessionID, toolName string) error — the actual limit check, split out so both servers can share one limiter and one budget.
2. OrgNameArgument and BuildAllowedOrganizationsMap — exported so the go-sdk allowlist check doesn't duplicate them.

- cmd/terraform-mcp-server: passed the existing rateLimiter down to the official-SDK server instead of creating a second one.

- The two new middlewares that log (OrganizationAllowlistMiddleware, ToolLoggingMiddleware) use slog. RateLimitMiddleware doesn't log directly — it delegates to the shared limiter, which keeps logging the way it already did.

## PCI review checklist

### DIFF
--- a/cmd/terraform-mcp-server/init.go
+++ b/cmd/terraform-mcp-server/init.go
@@ -17,6 +17,7 @@ import (
 
 	"github.com/hashicorp/terraform-mcp-server/pkg/client"
 	mcpofficial "github.com/hashicorp/terraform-mcp-server/pkg/mcp-official"
+	"github.com/hashicorp/terraform-mcp-server/pkg/mcp-official/tools/middleware"
 	"github.com/hashicorp/terraform-mcp-server/pkg/resources"
 	"github.com/hashicorp/terraform-mcp-server/pkg/tools"
 	"github.com/hashicorp/terraform-mcp-server/pkg/toolsets"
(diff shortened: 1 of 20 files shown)


cmd/terraform-mcp-server/init.go
```diff
@@ -458,13 +458,25 @@ func streamableHTTPServerInit(ctx context.Context, hcServer *server.MCPServer, l
 	return nil
 }
 
-func getOfficialStreamableServer(ctx context.Context, heartbeatInterval time.Duration, isStateless bool, corsConfig client.CORSConfig, logger *log.Logger, organizationAllowlist []string, enabledToolsets []string) http.Handler {
+func getOfficialStreamableServer(ctx context.Context, heartbeatInterval time.Duration, isStateless bool, corsConfig client.CORSConfig, logger *log.Logger, organizationAllowlist []string, enabledToolsets []string, rateLimiter *client.RateLimitMiddleware) http.Handler {
 	logger.Info("Creating a go-sdk StreamableHTTP server...")
-	hcServer := mcpofficial.NewServer(version.Version, instructions, heartbeatInterval, logger, enabledToolsets)
+	slogLogger := newSlogLogger(logger)
+	serverOpts := []mcpofficial.Option{
```
jaylonmcshan19-x: nitpick - could we build one middleware slice and pass it in a single `WithMiddlewares` call instead of two? It would prob make the ordering a bit easier to see too.

jaylonmcshan19-x (dismissed):

jrhouston (dismissed):
@MulyaP has started introducing a package structure here: https://github.com/hashicorp/terraform-mcp-server/commit/0b9d2d31cbc6c8c2dc429c4cacee601a0b5f22b1

For this change, I'd propose we have a new package for tool middlewares under `tools/middlware`

cmd/terraform-mcp-server/init.go
Shweta-Deshpande: Good idea @jaylonmcshan19-x I'll add that

Shweta-Deshpande:
> @MulyaP has started introducing a package structure here: [0b9d2d3](https://github.com/hashicorp/terraform-mcp-server/commit/0b9d2d31cbc6c8c2dc429c4cacee601a0b5f22b1)
> 
> For this change, I'd propose we have a new package for tool middlewares under `tools/middlware`

Done

jaylonmcshan19-x (approved):

PR #505, opened 2026-08-31
Migrate official MCP SDK logging from Logrus to slog

glbyktjys:
**This PR:**
- Initializes a slog logger for the official MCP server in HTTP mode, defaulting to stderr and preserve the existing `log-file` flag — NO longer writing to a separate `terraform-mcp-official.log`.
- Passes the logger to the official SDK for server and tool call logging.
-  Simplifies the tool-logging middleware to record successful calls, protocol-level errors, and tool-level errors (IsError) as structured slog records.

**What hasn’t changed:**
- Stdio logging remains on Logrus.
-  Shared pkg/client helpers continue using Logrus, since they're also used by the legacy mark3labs server.

## PCI review checklist

### DIFF
--- a/cmd/terraform-mcp-server/init.go
+++ b/cmd/terraform-mcp-server/init.go
@@ -40,7 +40,7 @@ type healthResponse struct {
 	Version   string `json:"version"`
 }
 
-const officialSlogOutputPath = "terraform-mcp-official.log"
+const officialSlogOutputFileName = "terraform-mcp-official.log"
 
 var (
 	rootCmd = &cobra.Command{
@@ -419,13 +419,13 @@ func streamableHTTPServerInit(ctx context.Context, hcServer *server.MCPServer, l
 		logger.Info("TF_X_OFFICIAL_SDK_ENABLED set to true in env, enabling the official mcp go-sdk server")
 		slogLevel := getSlogLevel(rootCmd)
 		slogFormat := getLogFormat(rootCmd)
-		officialLogger, officialLogFile, err := initSlog(officialSlogOutputPath, slogLevel, slogFormat)
+		officialLogger, officialLogFile, err := initSlog(officialSlogOutputFileName, slogLevel, slogFormat)
 		if err != nil {
-			return fmt.Errorf("failed to initialize official MCP slog logger at %s: %w", officialSlogOutputPath, err)
+			return fmt.Errorf("failed to initialize official MCP slog logger at %s: %w", officialSlogOutputFileName, err)
 		}
 		defer func() {
 			if err := officialLogFile.Close(); err != nil {
-				logger.Errorf("Failed to close official MCP log file at %s: %v", officialSlogOutputPath, err)
+				logger.Errorf("Failed to close official MCP log file at %s: %v", officialSlogOutputFileName, err)
 			}
 		}()
 
--- a/cmd/terraform-mcp-server/init.go
+++ b/cmd/terraform-mcp-server/init.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"io"
 	stdlog "log"
+	"log/slog"
 	"net/http"
 	"os"
 	"path"
@@ -38,6 +39,8 @@ type healthResponse struct {
 	Endpoint  string `json:"endpoint"`
 	Version   string `json:"version"`
 }
+
+const officialSlogOutputPath = "terraform-mcp-official.log"
 
 var (
 	rootCmd = &cobra.Command{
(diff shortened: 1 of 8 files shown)


pkg/mcp-official/tools/middleware/tool_logging.go
```diff
@@ -19,12 +19,42 @@ func ToolLogging(logger *slog.Logger) mcp.Middleware {
 				return next(ctx, method, req)
 			}
 
-			if params, ok := req.GetParams().(*mcp.CallToolParamsRaw); ok && logger != nil {
-				logger.InfoContext(ctx, "tool call executed",
-					"tool", params.Name, "arguments", string(params.Arguments))
+			toolName := "unknown"
+			arguments := ""
+			if params, ok := req.GetParams().(*mcp.CallToolParamsRaw); ok {
```
Shweta-Deshpande, line 24: I see the defensive check on `logger != nil` was removed here. Do we not need it anymore?

cmd/terraform-mcp-server/init.go
```diff
@@ -39,6 +40,8 @@ type healthResponse struct {
 	Version   string `json:"version"`
 }
 
+const officialSlogOutputPath = "terraform-mcp-official.log"
```
Shweta-Deshpande: Should this const be renamed to officialSlogOutputFileName instead of Path? I'm just wondering where it tries to look for this file in initSlog. Is it in the cwd?

jrhouston: How come we need to add this? The old code just grabs this from the `--log-file` CLI argument: https://github.com/hashicorp/terraform-mcp-server/blob/main/cmd/terraform-mcp-server/init.go#L141

pkg/mcp-official/tools/middleware/tool_logging.go
glbyktjys: Thanks for catching that! I missed preserving the nil guard while expanding the logging logic. I’ll restore it so the middleware continues without logging if the logger is nil.

cmd/terraform-mcp-server/init.go
glbyktjys: Agreed the new name would be clearer. It’s a relative filename, so the log is created in the cwd. If the server restarts from the same directory, it appends to the existing file. I’ll rename it accordingly.

cmd/terraform-mcp-server/init.go
```diff
@@ -39,6 +40,8 @@ type healthResponse struct {
 	Version   string `json:"version"`
 }
 
+const officialSlogOutputFileName = "terraform-mcp-official.log"
```
jaylonmcshan19-x: hmmm, I believe  `k8s` only picks up `stdout/stderr`, so anything written to
a file in the container won't show up in kubectl logs or even make its way to the instana collector I think.
From my understanding `LOG_FORMAT=json` exists so a collector can be able to parse our output and this way my sidestep it

Just curios was there a reason for the separate file over just writing to stdout with something like a component attr to be able to tell them apart? @jrhouston Can you confirm if my above statement is correct or not, you def have a better understanding than I when it comes to K8s

cmd/terraform-mcp-server/init.go
```diff
@@ -370,7 +417,20 @@ func streamableHTTPServerInit(ctx context.Context, hcServer *server.MCPServer, l
 	// Create the official go-sdk streamable server
 	if enableOfficialSDK := os.Getenv("TF_X_OFFICIAL_SDK_ENABLED"); enableOfficialSDK == "true" {
 		logger.Info("TF_X_OFFICIAL_SDK_ENABLED set to true in env, enabling the official mcp go-sdk server")
-		officialStreamableServer := getOfficialStreamableServer(ctx, heartbeatInterval, isStateless, corsConfig, logger, organizationAllowlist, enabledToolsets, rateLimiter)
+		slogLevel := getSlogLevel(rootCmd)
+		slogFormat := getLogFormat(rootCmd)
+		officialLogger, officialLogFile, err := initSlog(officialSlogOutputFileName, slogLevel, slogFormat)
+		if err != nil {
```
jaylonmcshan19-x, line 432: I wonder if this could poissibly break the base image. that stage is `FROM` scratch with no `WORKDIR`, so `cwd` is / and we run as `uid 65532` from looking into it. nothing in the filesystem but the binary and the ca bundle, and / is root owned, so `OpenFile` on a relative path will fail I think and the server won't start at all with this > `TF_X_OFFICIAL_SDK_ENABLED=true.` This will prob go away if we write to `stdout`. Also another thing to think about whether or not a log file failing should stop the server from starting, curios to hear your opinion :)

cmd/terraform-mcp-server/init.go
glbyktjys: yeah, I’m a bit unsure about this approach as well. The official MCP Go SDK currently shares the same HTTP server startup path as the existing mark3labs implementation, so it’s bit difficult to cleanly separate their logs during the migration. The separate file was intended as a temporary way to isolate the official SDK logs.

cmd/terraform-mcp-server/init.go
glbyktjys: yes, you’re right. The current OpenFile call will likely fail, and that failure would prevent the server from starting. And logging failure probably shouldn’t stop the server.
My main concern is avoiding changes to the existing mark3labs logger and startup behavior during the migration. Would it be better to revisit this once the official server startup is fully isolated? If separate ingestion is still needed, probably we will need explicit collector configuration then.
In the meantime, would you recommend any alternatives for keeping the official SDK logs distinguishable without affecting the existing mark3labs logging path?

jrhouston (approved):

PR #506, opened 2026-08-31
[Official SDK Migration] - add workspace crud tools

MulyaP:
## Summary

This PR implements official MCP Go SDK versions of four Terraform workspace tools:

- `create_workspace`
- `get_workspace_details`
- `update_workspace`
- `delete_workspace_safely`

These tools previously existed in the `mark3labs/mcp-go` implementation.

## Implementation Changes

The tools now use `github.com/modelcontextprotocol/go-sdk/mcp` and its typed tool registration API.

Compared with the legacy implementation:

- `create_workspace`, `update_workspace`, and `delete_workspace_safely` return concise mutation results containing:
  - Workspace ID
  - Workspace name
  - Operation result message
- Unsupported workspace tag updates are no longer exposed by `update_workspace`.
- `delete_workspace_safely` remains protected by the Terraform operations feature gate but under a new variable name called `PREVENT_DESTRUCTIVE_OPERATIONS`


pkg/mcp-official/tools/tfe/get_workspace_details.go
```diff
@@ -0,0 +1,83 @@
...
+	return &mcp.Tool{
+		Name:        "get_workspace_details",
+		Description: "Fetches detailed information about a specific Terraform workspace, including configuration, variables, and current state information.",
+		Annotations: &mcp.ToolAnnotations{Title: "Get detailed information about a Terraform workspace", OpenWorldHint: ptr(true), ReadOnlyHint: true, DestructiveHint: ptr(false)},
+	}
+}
+
+func GetWorkspaceDetailsFunc(ctx context.Context, request *mcp.CallToolRequest, input GetWorkspaceDetailsArguments) (*mcp.CallToolResult, any, error) {
+	orgName := strings.TrimSpace(input.TerraformOrgName)
+	workspaceName := strings.TrimSpace(input.WorkspaceName)
```
Marin2409, line 40: For input validation I like to split these into individual validations with specific error messages. Providing explicit error messages for each missing field helps the LLM debug its tool calls much faster than a combined error message.

```go
if orgName == "" {
    return nil, nil, fmt.Errorf("terraform_org_name is required")
}
if workspaceName == "" {
    return nil, nil, fmt.Errorf("workspace_name is required")
}
```

pkg/mcp-official/tools/tfe/delete_workspace_safely.go
```diff
@@ -0,0 +1,52 @@
...
+	if workspaceID == "" {
+		return nil, nil, fmt.Errorf("workspace_id is required")
+	}
+	tfeClient, err := client.GetTfeClient(ctx, client.SessionIDFromRequest(request))
+	if err != nil {
+		return nil, nil, err
+	}
+	if err != nil {
+		return nil, nil, err
+	}
```
Marin2409: I think you have the err check twice.

pkg/mcp-official/tools/tfe/get_workspace_details.go
```diff
@@ -0,0 +1,83 @@
...
+
+	var buf bytes.Buffer
+	if err := jsonapi.MarshalPayload(&buf, payload); err != nil {
+		return nil, nil, fmt.Errorf("failed to marshal workspace: %w", err)
+	}
+
+	return &mcp.CallToolResult{
+		Content: []mcp.Content{&mcp.TextContent{Text: buf.String()}},
+	}, nil, nil
+}
```
Marin2409, line 86: Another thing here: the manual `jsonapi.MarshalPayload` block isn't needed with the new SDK. The new go-sdk handler signature has three return values: (*mcp.CallToolResult, T, error). That middle typed return is the payload, the SDK picks it up and calls json.Marshal on it automatically. You never need to touch a buffer or build a CallToolResult yourself. 

What you can do is to define a new local response struct with plain json tags (similar to ProjectDetails in `projects_get_tool.go` See [#509](https://github.com/hashicorp/terraform-mcp-server/pull/509) for reference. ) instead of reusing `WorkspaceToolResponse` from `pkg/client/types.go`, that type is annotated for jsonapi and belongs to the old stack. Populate the new struct from the `*tfe.Workspace` you already have, and return it as the second value. 

Something like this:

```go
// Define a local response type with plain json tags (no jsonapi)
type GetWorkspaceDetailsResponse struct {
    WorkspaceID string          `json:"workspace_id"`
    Workspace   *tfe.Workspace  `json:"workspace"`
    Variables   []*tfe.Variable `json:"variables,omitempty"`
    Readme      string          `json:"readme,omitempty"`
}

func GetWorkspaceDetailsFunc(...) (*mcp.CallToolResult, *GetWorkspaceDetailsResponse, error) {
    // ... fetch workspace, variables, readme ...

    return nil, &GetWorkspaceDetailsResponse{
        WorkspaceID: workspace.ID,
        Workspace:   workspace,
        Variables:   workspaceVariables,
        Readme:      readme,
    }, nil
    // No bytes.Buffer, no jsonapi, no manual CallToolResult construction
}
```

pkg/mcp-official/tools/tfe/create_workspace.go
```diff
@@ -0,0 +1,143 @@
...
+		options.VCSRepo = &tfe.VCSRepoOptions{
+			Identifier:   &input.VCSRepoIdentifier,
+			OAuthTokenID: &input.VCSRepoOAuthTokenID,
+		}
+		if input.VCSRepoBranch != "" {
+			options.VCSRepo.Branch = &input.VCSRepoBranch
+		}
+	}
+
+	tfeClient, err := client.GetTfeClient(ctx, client.SessionIDFromRequest(request))
```
Marin2409, line 103: I think you have the err check twice

pkg/mcp-official/tools/tfe/create_workspace.go
```diff
@@ -0,0 +1,143 @@
...
+		},
+	}
+}
+
+func CreateWorkspaceFunc(ctx context.Context, request *mcp.CallToolRequest, input CreateWorkspaceArguments) (*mcp.CallToolResult, *WorkspaceMutationResult, error) {
+	orgName := strings.TrimSpace(input.TerraformOrgName)
+	workspaceName := strings.TrimSpace(input.WorkspaceName)
+	if orgName == "" || workspaceName == "" {
+		return nil, nil, fmt.Errorf("terraform_org_name and workspace_name are required")
+	}
```
Marin2409: For input validation I like to split these into individual validations with specific error messages. Providing explicit error messages for each missing field helps the LLM debug its tool calls much faster than a combined error message.

```go
if orgName == "" {
    return nil, nil, fmt.Errorf("terraform_org_name is required")
}
if workspaceName == "" {
    return nil, nil, fmt.Errorf("workspace_name is required")
}
```

pkg/mcp-official/tools/tfe/create_workspace.go
```diff
@@ -0,0 +1,143 @@
...
+
+	executionMode, err := parseExecutionMode(input.ExecutionMode, true)
+	if err != nil {
+		return nil, nil, err
+	}
```
Marin2409: Correct me if I'm wrong but the `execution_mode` default isn't doing what it looks like it's doing.

On line 61, `parseExecutionMode` is called with `allowEmpty: true`, so when the user doesn't pass an execution mode it hands back "remote" as the default. But the `if` on line 83 checks `input.ExecutionMode` (the original input) rather than the value that came back from the parser. So when the input is empty, that `if` is false, `options.ExecutionMode` never gets set, and the "remote" we just worked out is thrown away.

When ExecutionMode is left nil, TFE fills it in itself, from the project's default execution mode if a `project_id` was passed, otherwise from the organization's. So the tool advertises remote but actually creates the workspace with whatever that org or project happens to be set to. 

There's a smaller related problem too. Because line 83 checks the raw string, an input of "  " gets past it, then gets trimmed to "" inside the parser, hits the allowEmpty branch, and silently becomes remote. 

What you could maybe do, you can remove:

```go
executionMode, err := parseExecutionMode(input.ExecutionMode, true)
if err != nil {
	return nil, nil, err
}
```

And move the call inside the check on line 83, passing false:

```go
if input.ExecutionMode != "" {
	mode, err := parseExecutionMode(input.ExecutionMode, false)
	if err != nil {
		return nil, nil, err
	}
	options.ExecutionMode = &mode
}
```

This mirrors how you already have it in update_workspace.go (lines 73-79)

Marin2409 (changes requested):
Also, for file naming convention see how how Im doing it here #509 . Let me know what you think.

pkg/mcp-official/tools/tfe/create_workspace.go
Mulya-Patel: I agree, will change it.

pkg/mcp-official/tools/tfe/create_workspace.go
Mulya-Patel: Yeah, that's my bad😅

pkg/mcp-official/tools/tfe/get_workspace_details.go
Mulya-Patel: I actually did try this and it wouldn't work because types like `tfe.Workspaces` has some types defined in it (like `tfe.Workspaces.Link` uses type  map[string]interface{} which might be one of the culprits) which the SDK's json marshaller is not able to pick up and it ends up throwing an error. `projects_get_tool.go` response structure works because it is using all primitive data types and none of the tfe defined custom/problematic types. I actually did try to switch out your custom defined response struct with `tfe.Project` and I got the same error.

pkg/mcp-official/tools/tfe/delete_workspace_safely.go
Mulya-Patel: Yep, my bad again 😅

pkg/mcp-official/tools/tfe/get_workspace_details.go
Mulya-Patel: Agreed on that!

Mulya-Patel:
> Also, for file naming convention see how how Im doing it here #509 . Let me know what you think.

I believe keeping file names exactly the same as tool names would make it easier for any developer to navigate this. But again, its a personal opinion and I can be wrong.

pkg/mcp-official/tools/tfe/create_workspace.go
Mulya-Patel: That's a good catch @Marin2409. I will make these changes right now, thanks for pointing it out.

pkg/mcp-official/tools/tfe/get_workspace_details.go
Marin2409: You're right, my example was wrong. I suggested embedding `tfe.Workspace` directly, which causes the exact same panic. The SDK walks the entire output type at startup to generate the JSON Schema, and `tfe.Workspace` has fields like `Links map[string]interface{}` and some polyrelation union types that the schema generator can't represent.

Another option is to add a fully flat struct with only primitive fields, with no raw `go-tfe` types anywhere in the struct. I went ahead and applied that approach in the branch just to test it out. Here's what it would look like:

```go
type WorkspaceVariable struct {
    Key         string `json:"key"`
    Value       string `json:"value,omitempty"`
    Description string `json:"description,omitempty"`
    Category    string `json:"category"`
    Sensitive   bool   `json:"sensitive"`
    HCL         bool   `json:"hcl"`
}

type WorkspaceGetToolResponse struct {
	WorkspaceID          string              `json:"workspace_id"`
	WorkspaceName        string              `json:"workspace_name"`
	OrganizationName     string              `json:"organization_name,omitempty"`
	ProjectID            string              `json:"project_id,omitempty"`
	Description          string              `json:"description,omitempty"`
	ExecutionMode        string              `json:"execution_mode,omitempty"`
	TerraformVersion     string              `json:"terraform_version,omitempty"`
	WorkingDirectory     string              `json:"working_directory,omitempty"`
	AutoApply            bool                `json:"auto_apply"`
	Locked               bool                `json:"locked"`
	ResourceCount        int                 `json:"resource_count"`
	TagNames             []string            `json:"tag_names,omitempty"`
	TriggerPrefixes      []string            `json:"trigger_prefixes,omitempty"`
	TriggerPatterns      []string            `json:"trigger_patterns,omitempty"`
	SpeculativeEnabled   bool                `json:"speculative_enabled"`
	GlobalRemoteState    bool                `json:"global_remote_state"`
	CreatedAt            time.Time           `json:"created_at"`
	UpdatedAt            time.Time           `json:"updated_at"`
	Variables            []WorkspaceVariable `json:"variables,omitempty"`
	Readme               string              `json:"readme,omitempty"`
}
```

Then in the handler, map fields explicitly from *tfe.Workspace and loop over vars.Items to populate []WorkspaceVariable. No raw go-tfe types anywhere in the response tree, that's what lets the SDK introspect it cleanly.

I actually like the flat struct approach here. I know the manual field mapping is a bit more work upfront, but you get a cleaner output format, better compatibility with the SDK, and a more stable response contract.

`MarshalPayload` definitely makes things easier from a mapping perspective, so I understand the appeal (and the pain of having to map everything manually 😅). I just think the flat struct is a better fit for what the SDK expects.

Let me know what you think.

PR #508, opened 2026-09-01
Fix session-scoped TFE/HTTP client caching

Shweta-Deshpande:
## What are we solving?

- The official SDK transport was using a single global TFE client for every session — the first user's token got cached once and then reused for everyone else's requests until the server restarted. 
- The HTTP client had a similar issue: in stateless mode, every request reused one cached client instead of getting its own.

## What changed?

- This PR moves the caching/invalidation logic into pkg/client so both the old (mark3labs) and new (official SDK) transports share the same per-session cache. 
- Each session gets its own TFE/HTTP client, a token change on a session correctly replaces its cached client, and stateless requests never share a cached client with each other. 
- The official SDK's client code is now just a thin wrapper that calls into this shared logic.
- Added unit tests covering per-session caching, token rotation, and the stateless case for both transports.

## PCI review checklist

### DIFF
--- a/pkg/mcp-official/tools/tfe/organizations.go
+++ b/pkg/mcp-official/tools/tfe/organizations.go
@@ -43,8 +43,12 @@ func ListTerraformOrganizationsTool() *mcp.Tool {
 }
 
 func ListTerraformOrganizationsFunc(ctx context.Context, request *mcp.CallToolRequest, input ListOrganizationsArguments) (*mcp.CallToolResult, *OrganizationSummaryList, error) {
+	var sessionID string
+	if request.Session != nil {
+		sessionID = request.Session.ID()
+	}
 
-	tfeClient, err := client.GetTfeClient(ctx)
+	tfeClient, err := client.GetTfeClient(ctx, sessionID)
 	if err != nil {
 		return nil, nil, err
 	}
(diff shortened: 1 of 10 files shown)


pkg/mcp-official/tools/tfe/organizations.go
```diff
@@ -43,8 +43,12 @@ func ListTerraformOrganizationsTool() *mcp.Tool {
 }
 
 func ListTerraformOrganizationsFunc(ctx context.Context, request *mcp.CallToolRequest, input ListOrganizationsArguments) (*mcp.CallToolResult, *OrganizationSummaryList, error) {
+	var sessionID string
```
jaylonmcshan19-x: with more tools migrating we're gonna prob keep repeating this nil check in every tool func. wondering if a small helper in pkg/client would be useful here?

jaylonmcshan19-x (dismissed):

pkg/mcp-official/tools/tfe/organizations.go
Shweta-Deshpande: Yes, sounds good. Done.

Shweta-Deshpande: Added the helper in mcp-official since *mcp.CallToolRequest is specific to the official sdk

jaylonmcshan19-x (approved):

PR #509, opened 2026-09-01
[Official SDK Migration] - Project tools

Marin2409:
This PR migrates `get_project`, `create_project`, `delete_project`, and `list_terraform_projects` from `mark3labs/mcp-go` to the official `modelcontextprotocol/go-sdk`.

New tools (`pkg/mcp-official/tools/tfe/`):
`projects_get_tool.go`, `projects_create_tool.go`, `projects_delete_tool.go`, `projects_list_tool.go`

It also adds `pagination.go` — shared abstractions across all list tools:

- `Pagination` embed and `ListOptions()` for normalising page/pageSize inputs
- `PaginationDetails` value type replacing direct `*tfe.Pagination` embedding — `tfe.Pagination` fields have no `omitempty`, so the SDK treated them all as required output, causing validation failures when pagination was nil
- `paginationSchemaProperties()` — reusable `page/pageSize` property map with `minimum/maximum` bounds for `input schemas`
- `ptr` utility

Notes:

- `Error handling` — handlers return `fmt.Errorf` directly. The SDK sets `isError: true` and surfaces the message to the client automatically, so no custom error helper or logger threading is needed. See the [SDK reference](https://pkg.go.dev/github.com/modelcontextprotocol/go-sdk/mcp#ToolHandlerFor) and the [implementation](https://github.com/modelcontextprotocol/go-sdk/blob/bc72835f62eb94d0fb484439f886b6885b075f36/mcp/protocol.go#L302-L304).
- `tools.go` — tools are registered via an `officialFactories` map keyed by tool name. `projects_delete_tool.go` gated behind `ENABLE_TF_OPERATIONS` 
- `list` tools (`projects_list_tool.go`, `workspaces_list_tool`, and `organizations_list_tool`) return `{"items": []}` when a page holds no results. `list_terraform_orgs` and `list_workspaces` previously returned a tool error, in both the legacy and migrated implementations; `list_terraform_projects` already returned an empty array. An empty page means the call succeeded and matched nothing, so reporting it as a failure tells the model the tool broke when it worked.  See [Official MCP spec](https://modelcontextprotocol.io/specification/2026-07-28/server/tools#error-handling)

Flags: 

- `create_project` drops `agent`. Both implementations (SDK and Legacy) advertised `agent`, and neither accepts `default_agent_pool_id`, which `TFC` requires for that mode — so the call always failed at the API. The enum is now `local`, `remote`.

Next objective, implement Agent Pool management tools:

- `create_agent_pool` (`agent_pools_create_tool.go`) — creates an agent pool. Pools are organization-scoped, not project-scoped: `AgentPools.Create` takes an organization, and `AgentPoolCreateOptions` exposes `OrganizationScoped` plus `AllowedProjects`/`AllowedWorkspaces` for narrowing a pool to specific resources.

## PCI review checklist


pkg/mcp-official/tools/tools.go
```diff
@@ -1,18 +1,44 @@
...
+
+	tfeTools "github.com/hashicorp/terraform-mcp-server/pkg/mcp-official/tools/tfe"
 	"github.com/hashicorp/terraform-mcp-server/pkg/toolsets"
+	"github.com/hashicorp/terraform-mcp-server/pkg/utils"
 	"github.com/modelcontextprotocol/go-sdk/mcp"
 	log "github.com/sirupsen/logrus"
-	tfeTools "github.com/hashicorp/terraform-mcp-server/pkg/mcp-official/tools/tfe"
 )
 
+func areDestructiveOperationsPrevented() bool {
```
jaylonmcshan19-x: Wondering if this is backwards? it returns true when `PREVENT_DESTRUCTIVE_OPERATIONS` is` "false"`,and then the call site below registers `delete_project` when it's` true`. so `delete_project`
shows up when destructive ops are prevented and disappears when they're allowed. Wondering if the name or the comparison needs flipping

pkg/mcp-official/tools/tfe/projects_get_tool.go
```diff
@@ -0,0 +1,85 @@
...
+	DefaultAgentPoolName        string `json:"default_agent_pool_name,omitempty"`
+}
+
+// GetProjectArguments holds the input parameters for fetching a single project.
+type GetProjectArguments struct {
+	// Required field
+	ProjectID string `json:"project_id" jsonschema:"The ID of the project to fetch (e.g., 'prj-abc123def456')"`
+}
+
+func GetProjectTool() *mcp.Tool {
```
jaylonmcshan19-x, line 34: `get_project` and d`elete_project` don't set `InputSchema` or `OutputSchema` but the list tools
and `create_project `do. wondering if its intended or just not needed for these?

Marin2409: In this case its intentional `get_project` and `delete_project` take a single `project_id` string with no extra constraints, e.g. no length limits, no enums, no pagination bounds. The SDK already infers the same schema automatically, so explicitly setting `InputSchema` adds nothing (in this case). Same for `OutputSchema`, both return simple flat structs with no pointer slices, so the SDK's auto-inferred output schema has no nullable type arrays to worry about. No warning, no need to override.

The `List tools` and `create_project` need explicit schemas because:

- Without the `InputSchema` the SDK infers a basic schema from the struct tags, field names, types, and which are required. That's fine for simple tools. But for `create_project`, the schema the client sees would have no `minLength`/`maxLength` on `project_name`, no `maxLength` on `description`, no `enum` on `default_execution_mode`. 

- Without `OutputSchema` the SDK still auto-generates one from the return type. For simple structs like `DeleteProjectResponse` or `ProjectDetails` that's fine. For `list tools` returning `[]*T`, you'd get `type: ["null", "array"]` in the advertised schema, which breaks MCP clients that read type as a scalar string.

- Without `narrowNullableTypes`, setting `OutputSchema: inferSchema[T]` directly would still emit the broken `["null", "array"]`. It's `outputSchema[T]` (which wraps `inferSchema` + `narrowNullableTypes`) that actually fixes it, not just setting `OutputSchema` to anything.

pkg/mcp-official/tools/tools.go
Marin2409: You are right, good catch. Its a bit confusing. I updated the function's name to `areDestructiveOperationsAllowed`:

```go
func areDestructiveOperationsAllowed() bool {
	envVar := utils.GetEnv("PREVENT_DESTRUCTIVE_OPERATIONS", "true")
	return strings.ToLower(envVar) == "false"
}
```

Now: 

If `PREVENT_DESTRUCTIVE_OPERATIONS=true`, then `func areDestructiveOperationsAllowed()` returns `false`, therefore destruction operations are **not allowed**. 

If `PREVENT_DESTRUCTIVE_OPERATIONS=false`, then `func areDestructiveOperationsAllowed()` returns `true`, therefore destruction operations are **allowed**. 

Let me know if its still confusing lol

jaylonmcshan19-x: that reads much better, thank ya :)
 is `PREVENT_DESTRUCTIVE_OPERATIONS` meant to replace `ENABLE_TF_OPERATIONS` once we're fully off mark3labs? asking because the helm chart PR I made exposes ENABLE_TF_OPERATIONS as a val, and if it's getting renamed I will sort that out now in my PR

jaylonmcshan19-x (dismissed):

pkg/mcp-official/tools/tools.go
Marin2409: Good call, let's hold the rename until the migration is complete. Changing `ENABLE_TF_OPERATIONS` now would break the legacy tools and the `HCPT tests` that depend on it. I'll handle the rename in a separate PR after the full migration to avoid conflicts in the OSS repo.

pkg/mcp-official/tools/tools.go
```diff
@@ -1,18 +1,44 @@
...
+
+	tfeTools "github.com/hashicorp/terraform-mcp-server/pkg/mcp-official/tools/tfe"
 	"github.com/hashicorp/terraform-mcp-server/pkg/toolsets"
+	"github.com/hashicorp/terraform-mcp-server/pkg/utils"
 	"github.com/modelcontextprotocol/go-sdk/mcp"
 	log "github.com/sirupsen/logrus"
-	tfeTools "github.com/hashicorp/terraform-mcp-server/pkg/mcp-official/tools/tfe"
 )
 
+func areDestructiveOperationsAllowed() bool {
```
jrhouston: I would name this `destructiveOperationsEnabled()`

pkg/mcp-official/tools/tools.go
```diff
@@ -1,18 +1,44 @@
...
+	envVar := utils.GetEnv("PREVENT_DESTRUCTIVE_OPERATIONS", "true")
```
jrhouston: I think we're going to stick with the current env var name for now? Changing it will happen as a breaking change in a different PR

pkg/mcp-official/tools/tfe/workspaces_list_tool.go
```diff
@@ -0,0 +1,164 @@
...
+		return nil, &WorkspaceSummaryList{
+			Items:             nonNilSlice(summaries),
+			PaginationDetails: paginationDetails(workspaces.Pagination),
+		}, nil
+	}
+}
+
+// joinTrimmedCSV normalizes a comma-separated list by trimming whitespace around
+// each element, so "a, b , c" is sent to the API as "a,b,c".
+func joinTrimmedCSV(csv string) string {
```
jrhouston: Go has an `encoding/csv` package you can probably use for this.

pkg/mcp-official/tools/tfe/projects_delete_tool.go
```diff
@@ -0,0 +1,73 @@
...
+			Type: "object",
+			Properties: map[string]*jsonschema.Schema{
+				"project_id": {
+					Type:        "string",
+					Description: "The ID of the project to delete (e.g., 'prj-abc123def456')",
+				},
+			},
+			PropertyOrder:        []string{"project_id"},
+			Required:             []string{"project_id"},
+			AdditionalProperties: &jsonschema.Schema{Not: &jsonschema.Schema{}},
```
jrhouston: Why do we need to create the JSON schema for this one if it's just one required string parameter? Can't we just infer the schema?

jrhouston (changes requested):
Couple things

pkg/mcp-official/tools/tools.go
Marin2409: Agreed. Kept with legacy env var. Will change after migration.

pkg/mcp-official/tools/tools.go
Marin2409: Agreed. func name updated

pkg/mcp-official/tools/tfe/projects_delete_tool.go
Marin2409: Yes, we can infer schema in this case. As discussed before, better to build schemas when dealing with complex arguments. For smaller schemas easier to infer.

pkg/mcp-official/tools/tfe/workspaces_list_tool.go
Marin2409: I dont think `encoding/csv` would be the best option here since that package parses and writes csv files. here its just normalizing a comma separate list. I changed the name of the function to `normaliseCommaSeparated` a little bit less confusing.

pkg/mcp-official/tools/tools.go
```diff
@@ -18,11 +18,23 @@ var officialFactories = map[string]func(svr *mcp.Server){
...
+	"list_terraform_projects": func(svr *mcp.Server) {
+		mcp.AddTool(svr, tfeTools.ListProjectsTool(), tfeTools.ListProjectsFunc)
+	},
+	"create_project": func(svr *mcp.Server) {
+		mcp.AddTool(svr, tfeTools.CreateProjectTool(), tfeTools.CreateProjectFunc)
+	},
+	"get_project": func(svr *mcp.Server) {
+		mcp.AddTool(svr, tfeTools.GetProjectTool(), tfeTools.GetProjectFunc)
+	},
+	"delete_project": func(svr *mcp.Server) {
```
jaylonmcshan19-x, line 38: This is a lot better with the per-tool tests!!!

one thing though, looks like the destructive ops gate dropped out if im not mistaken. `areDestructiveOperationsAllowed` isn't in the diff anymore from what I can see and `delete_project` is sitting in `officialFactories` with nothing around it, so it registers whenever the toolset is on. guessing it got lost rebasing onto Shweta's toolfilter work?

Also `RegisterTools` loops through `AllTools` and checks` filter.IsToolEnabled` but not `td.RequiresTFOps`, and in`registry.go` has `delete_project` marked as `RequiresTFOps: true`. I think adding that check to the loop would cover this one and anything destructive that migrates after it.

Marin2409: Yea, it was lost when rebased. And good catch with `td.RequiresTFOps`. Check Added!  Thanks Jaylon.

pkg/mcp-official/tools/tfe/pagination.go
```diff
@@ -0,0 +1,86 @@
...
+			Maximum:     ptr(float64(maxPageSize)),
+		},
+	}
+}
+
+// ptr is a convenience helper for taking the address of a literal value,
+// needed wherever the SDK or schema types expect a *bool, *float64, etc.
+func ptr[T any](v T) *T {
+	return &v
+}
```
glbyktjys: One nit: since we already depend directly on jsonschema, shall we use `jsonschema.Ptr()` and remove the local ptr helper?

pkg/mcp-official/tools/tfe/projects_list_tool.go
```diff
@@ -0,0 +1,94 @@
...
+type ProjectSummary struct {
+	ID   string `json:"project_id"`
+	Name string `json:"project_name"`
+}
+
+// ProjectSummaryList is a list of project summaries and pagination details
+type ProjectSummaryList struct {
+	Items []*ProjectSummary `json:"items"`
+	PaginationDetails
+}
```
glbyktjys, line 27: I was seeing a schema warning in MCP Inspector for state list tools: Items []*StateVersionSummary is inferred with type: ["null", "array"], and its elements with type: ["null", "object"]. 
Inspector warns that some agent clients expect type to be a single string rather than an array.
Not sure if you have the same warning in your end. Would it make sense to restore an explicitly normalized OutputSchema, or is this warning safe to ignore for now?

pkg/mcp-official/tools/tfe/pagination.go
Marin2409: Niceee, I hadn't seen`jsonschema.Ptr()`. Tools refactored to jsonschema.Ptr()

pkg/mcp-official/tools/tfe/projects_list_tool.go
Marin2409: yeah, Inspector may warn when Go schema inference produces nullable unions from slice and pointer types. For example, `[]*StateVersionSummary` alone can infer a root type of `["null", "array"]`, with `["null", "object"]` for its items.

Under the current MCP specification, that schema is valid: `outputSchema` may be any valid JSON Schema 2020-12 document, and a tool’s `structuredContent` may be an object, array, primitive, or `null`, as long as it conforms to that output schema. See [JSON Schema 2020-12: `type`](https://json-schema.org/draft/2020-12/json-schema-validation#section-6.1.1)

For our object-rooted list outputs:

```go
type ProjectSummaryList struct {
    Items []*ProjectSummary `json:"items"`
    PaginationDetails
}
```

the tool schema root is still an object. The Go SDK dereferences a top-level pointer output before schema inference, and a struct is inferred as an object, see [`server.go` lines 467–472](https://github.com/modelcontextprotocol/go-sdk/blob/v1.7.0/mcp/server.go#L467-L472) and [`infer.go` lines 245–246](https://github.com/google/jsonschema-go/blob/v0.4.3/jsonschema/infer.go#L245-L246). That makes this shape more compatible with clients that assume object-shaped tool outputs.

but still, if we want the strongest contract and more compatibility, I think we should normalize list results over time: return an empty array rather than null, avoid null entries in items, and explicitly publish a schema with "type": "array" for items and "type": "object" for the item schema.

here are some useful links that i've found:
- https://blog.modelcontextprotocol.io/posts/2026-07-28-release-candidate/#full-json-schema-2020-12-for-tools
- https://modelcontextprotocol.io/seps/1613-establish-json-schema-2020-12-as-default-dialect-f

jaylonmcshan19-x:
@Marin2409 the `RequiresTFOps` gate via` filter.IsTFOperationsEnabled` is a lot cleaner than the env read nice!!

one thing I want to check, `narrowNullableTypes` and `nonNilSlice` are gone from what I can tell. the original description I believe said list outputs would emit ["null", "array"] without them and some clients don't handle that. did that turn out not to be a problem, or did it get lost in the pagination refactor? Only asking for context because my stacks PR copied the pattern and I'd rather match whatever you settled on lol 👍

Marin2409:
> @Marin2409 the `RequiresTFOps` gate via` filter.IsTFOperationsEnabled` is a lot cleaner than the env read nice!!
> 
> one thing I want to check, `narrowNullableTypes` and `nonNilSlice` are gone from what I can tell. the original description I believe said list outputs would emit ["null", "array"] without them and some clients don't handle that. did that turn out not to be a problem, or did it get lost in the pagination refactor? Only asking for context because my stacks PR copied the pattern and I'd rather match whatever you settled on lol 👍

Thanks Jaylon, and yea `narrowNullableTypes` and `nonNilSlice` are no longer implemented. They were not lost in a rebase, they were an alternative solution I was trying to solve the problem Jiayue and I were discussing here: https://github.com/hashicorp/terraform-mcp-server/pull/509#discussion_r4065942456

jaylonmcshan19-x:
> > @Marin2409 the `RequiresTFOps` gate via` filter.IsTFOperationsEnabled` is a lot cleaner than the env read nice!!
> > one thing I want to check, `narrowNullableTypes` and `nonNilSlice` are gone from what I can tell. the original description I believe said list outputs would emit ["null", "array"] without them and some clients don't handle that. did that turn out not to be a problem, or did it get lost in the pagination refactor? Only asking for context because my stacks PR copied the pattern and I'd rather match whatever you settled on lol 👍
> 
> Thanks Jaylon, and yea `narrowNullableTypes` and `nonNilSlice` are no longer implemented. They were not lost in a rebase, they were an alternative solution I was trying to solve the problem Jiayue and I were discussing here: [#509 (comment)](https://github.com/hashicorp/terraform-mcp-server/pull/509#discussion_r4065942456)

gotcha, that makes sense. I'll pull them out of the stacks PR so we match 👍 

one thing I've been thinking about, instead of walking the schema after the fact we could possibly just not return pointer slices. `[]StackSummary` instead of `[]*StackSummary` will infer as a plain array with object items, no null union at all. the summaries are just tiny structs so I don't think there'll be no real cost. same can likely apply for Items on the list types. With this, the inspector warning will go away without touching the schema. I can try it on stacks first and see if it holds up, but what do you think?

Marin2409:
> > > @Marin2409 the `RequiresTFOps` gate via` filter.IsTFOperationsEnabled` is a lot cleaner than the env read nice!!
> > > one thing I want to check, `narrowNullableTypes` and `nonNilSlice` are gone from what I can tell. the original description I believe said list outputs would emit ["null", "array"] without them and some clients don't handle that. did that turn out not to be a problem, or did it get lost in the pagination refactor? Only asking for context because my stacks PR copied the pattern and I'd rather match whatever you settled on lol 👍
> > 
> > 
> > Thanks Jaylon, and yea `narrowNullableTypes` and `nonNilSlice` are no longer implemented. They were not lost in a rebase, they were an alternative solution I was trying to solve the problem Jiayue and I were discussing here: [#509 (comment)](https://github.com/hashicorp/terraform-mcp-server/pull/509#discussion_r4065942456)
> 
> gotcha, that makes sense. I'll pull them out of the stacks PR so we match 👍
> 
> one thing I've been thinking about, instead of walking the schema after the fact we could possibly just not return pointer slices. `[]StackSummary` instead of `[]*StackSummary` will infer as a plain array with object items, no null union at all. the summaries are just tiny structs so I don't think there'll be no real cost. same can likely apply for Items on the list types. With this, the inspector warning will go away without touching the schema. I can try it on stacks first and see if it holds up, but what do you think?

yea you can try that but the null union will still be there regardless. I dug into the [jsonschema-go source](https://github.com/google/jsonschema-go/blob/794ce5e429b22f05aee47a62f9151c1e7b54fd22/jsonschema/infer.go#L55) and the ["null", "array"] is set unconditionally for any [reflect.Slice](https://github.com/google/jsonschema-go/blob/794ce5e429b22f05aee47a62f9151c1e7b54fd22/jsonschema/infer.go#L219), regardless of element type.

When jsonschema-go walks OrganizationSummaryList, it hits the Items field, sees its type is []*OrganizationSummary, and sets ["null", "array"] for the slice itself. Then it checks on the element type *OrganizationSummary, which follows the pointer, sets allowNull = true, and produces {"type": ["null", "object"]} for each item.

If we switch to []OrganizationSummary (value slice, value elements):

- The slice-level ["null", "array"] stays 
- The element-level ["null", "object"] goes away, no pointer to dereference, and items infers as plain {"type": "object"}

So the inspector warning you see on items being ["null", "array"] is from the slice kind. The secondary ["null", "object"] on each element is from the pointer element type. Switching to value slice fixes the second one, not the first.

The warning is cosmetic anyway. The ["null", "array"] on a nested items property is valid schema,  the [spec](https://blog.modelcontextprotocol.io/posts/2026-07-28-release-candidate/#full-json-schema-2020-12-for-tools) (also [see](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2106#issue-3825897254)) only restricts the root to be "object" for inputSchemas, which our wrapper structs already satisfy. Output schemas are unrestricted. 

That's why I suspect the inspector is warning about a style/convention mismatch, not a spec violation. The schema is valid, the tool works correctly.

PR #512, opened 2026-09-02
[Official SDK Migration] - User management tools

Marin2409:
This PR migrates `whoami` and `get_token_permissions` from `mark3labs/mcp-go` to the official `modelcontextprotocol/go-sdk`.

New tools (`pkg/mcp-official/tools/tfe/`):

- `user_whoami_tool.go` — returns the username, email, and service account flag for the currently authenticated token. 
- `user_get_token_permissions_tool.go` — fetches the human-readable list of permissions the current token holds in a given Terraform organization. 

Test:
- `user_whoami_tool_test.go` 
- `user_get_token_permissions_tool_test.go`

Also includes:

- `results.go` — shared `textResult()` helper that wraps a pre-serialized string into a `*mcp.CallToolResult`, used by tools that manage their own serialization instead of returning a typed struct.

## PCI review checklist


pkg/mcp-official/tools/tfe/user_whoami_tool.go
```diff
@@ -0,0 +1,54 @@
...
+
+// AccountDetails is the response shape returned by the whoami tool.
+type AccountDetails struct {
+	Username         string `json:"username"`
+	Email            string `json:"email"`
+	IsServiceAccount bool   `json:"is_service_account"`
+}
+
+// WhoAmIArguments holds the (empty) input for the whoami tool.
+type WhoAmIArguments struct{}
```
jrhouston: I think we don't need this, and in the handler function you can just use `_ any` as the params argument.

pkg/mcp-official/tools/tfe/user_whoami_tool.go
```diff
@@ -0,0 +1,54 @@
...
+			ReadOnlyHint:    true,
+			DestructiveHint: ptr(false),
+		},
+	}
+}
+
+func WhoAmIFunc(ctx context.Context, request *mcp.CallToolRequest, _ WhoAmIArguments) (*mcp.CallToolResult, *AccountDetails, error) {
+	tfeClient, err := client.GetTfeClient(ctx, client.SessionIDFromRequest(request))
+	if err != nil {
+		return nil, nil, fmt.Errorf("getting Terraform client: %w", err)
```
jrhouston, line 37: By doing this I think we're losing the value of what `ToolError` was doing in the previous implementation: https://github.com/hashicorp/terraform-mcp-server/blob/main/pkg/tools/tfe/whoami.go#L41

You can see what this function actually does is _not_ return an `error`:

https://github.com/hashicorp/terraform-mcp-server/blob/0df1ce70f00bf9b0bb7cca30dbc4b9df94abd1be/pkg/tools/tfe/errors.go#L13-L22

Instead it uses `NewToolResultError` which actually returns ToolResult but sets `IsError` to `true`: 

https://github.com/mark3labs/mcp-go/blob/51930cffc30e4708c60e9763a23ad0ed6ad06387/mcp/utils.go#L418-L428

jrhouston (changes requested):
Couple things

pkg/mcp-official/tools/tfe/user_whoami_tool.go
Marin2409: Good call. Struct removed and handler updated.

pkg/mcp-official/tools/tfe/user_whoami_tool.go
jrhouston: Nevermind – looks like we don't need to do this as the new SDK does this for us:

https://github.com/modelcontextprotocol/go-sdk/blob/bc72835f62eb94d0fb484439f886b6885b075f36/mcp/protocol.go#L302-L304

PR #514, opened 2026-09-03
[Official SDK Migration] - Add private registry tools

MulyaP:
This PR re-implements the following tools using official go-sdk for MCP:

- get_private_module_details
- get_private_provider_details
- search_private_modules
- search_private_providers

get_private_module_details implementation was referred from this [PR](https://github.com/hashicorp/terraform-mcp-server/pull/495)

Question: Is there a reason why we're building huge string outputs rather than returning json response like other tools?


Marin2409:
Private Registry tools migration has been continued in [PR #540](https://github.com/hashicorp/terraform-mcp-server/pull/540).

PR #517, opened 2026-09-08
[Official SDK Migration] Port runs tools to modelcontextprotocol/go-sdk

jrhouston:
## PCI review checklist

- [ ] I have documented a clear reason for, and description of, the change I am making.

- [ ] If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request.

- [ ] If applicable, I've documented the impact of any changes to security controls.

  Examples of changes to security controls include using new access control methods, adding or removing logging pipelines, etc.

PR #518, opened 2026-09-08
migrate state version tools

glbyktjys:
Migrate the state version tools from `mark3labs/mcp-go` to the official MCP Go SDK:

- `list_state_versions`
- `get_state_version`

## PCI review checklist


pkg/mcp-official/tools/tfe/state_version.go
```diff
@@ -0,0 +1,152 @@
...
+}
+
+// StateVersionDetails holds the complete JSON:API state version payload.
+type StateVersionDetails map[string]any
+
+type ListStateVersionsArguments struct {
+	TerraformOrgName string `json:"terraform_org_name" jsonschema:"The Terraform organization name"`
+	WorkspaceName    string `json:"workspace_name" jsonschema:"The workspace name to list state versions for"`
+	Page             int    `json:"page,omitempty" jsonschema:"Page number for pagination (min 1)"`
+	PageSize         int    `json:"pageSize,omitempty" jsonschema:"Results per page for pagination (min 1, max 100)"`
```
Marin2409, line 42: nit: for list tools I'd build the schema since the client wont be able to read your pagination constraints, e.g. (min 1) or (min 1, max 100). See: https://github.com/hashicorp/terraform-mcp-server/pull/517/changes#diff-f6bd0ca4a5643e85ae6a069eac6a41fa9b3784f85f27929dbf4f6db7b5ae5437R124-R133.

I also handle that [here](https://github.com/hashicorp/terraform-mcp-server/pull/509/changes#diff-5d781c6703fdecd0f59d221fc2722750034e258d8e154d467e70886cd4103e8fR39) , but you might have to wait until that one gets merge.

Marin2409 (changes requested):
LGTM! One small nit: maybe consider using one file per tool for better readability.

pkg/mcp-official/tools/tools.go
```diff
@@ -24,6 +24,12 @@ var officialFactories = map[string]func(svr *mcp.Server){
 	"list_terraform_orgs": func(svr *mcp.Server) {
 		mcp.AddTool(svr, tfeTools.ListTerraformOrganizationsTool(), tfeTools.ListTerraformOrganizationsFunc)
 	},
+	"list_state_versions": func(svr *mcp.Server) {
+		mcp.AddTool(svr, tfeTools.ListStateVersionsTool(), tfeTools.ListStateVersionsFunc)
+	},
+	"get_state_version": func(svr *mcp.Server) {
+		mcp.AddTool(svr, tfeTools.GetStateVersionTool(), tfeTools.GetStateVersionFunc)
```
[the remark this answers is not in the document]
Marin2409: If you wanna test tools against then official endpoint set `TF_X_OFFICIAL_SDK_ENABLED=true` and `TF_MCP_ENDPOINT="http://localhost:8080/mcp/official"`

But that's only if you wanna do it manually, on the CI is only wired to the legacy endpoint `/mcp`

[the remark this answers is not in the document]
glbyktjys: Yeah, I think it's good for now. I've manually tested it locally against the official MCP endpoint, and that should be sufficient for the current migration.

Marin2409 (approved):
LGTM! :)

pkg/mcp-official/tools/tfe/list_state_versions.go
```diff
@@ -0,0 +1,123 @@
...
+		Organization: terraformOrgName,
+		Workspace:    workspaceName,
+		ListOptions:  listOptions,
+	})
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to list workspace state versions: %w", err)
+	}
+	if len(stateVersions.Items) == 0 {
+		return nil, nil, fmt.Errorf("workspace has no state versions to list")
+	}
```
Marin2409, line 108: Actually, just one more thing I realized yesterday. Let me know what you think.

Whenever we return an error, the SDK automatically sets `IsError` to `true`, so the result gets treated as a tool error. See the [SDK reference](https://pkg.go.dev/github.com/modelcontextprotocol/go-sdk/mcp#ToolHandlerFor) and the [implementation](https://github.com/modelcontextprotocol/go-sdk/blob/bc72835f62eb94d0fb484439f886b6885b075f36/mcp/protocol.go#L302-L304).

I was doing the same thing for `list` tools, like in [#509](https://github.com/Marin2409/terraform-mcp-server/blob/2a9ddf685b51ab256f979375448186c6cbe6906a/pkg/mcp-official/tools/tfe/projects_list_tool.go#L81).

The thing I was thinking about is that if we check `len(...Items) == 0` and return an error message, the LLM could interpret that as the tool actually failing. But in this case, the tool succeeded, there just aren't any items to return.

So I ended up dropping the empty-check entirely, returning [] instead. The model can infer from an empty array that there's nothing to show, without thinking the tool broke

I don't think having no state versions, projects, or workspaces, etc, to list should be considered an error. It's a valid result of the query, there just happens to be nothing to list.

Marin2409 (changes requested):

pkg/mcp-official/tools/tfe/list_state_versions.go
glbyktjys: yes I agree probably just return an empty array with nil error and maybe have the "workspace has no state versions to list" in sturctured content like this:

```
var result *mcp.CallToolResult
if len(summaries) == 0 {
	result = &mcp.CallToolResult{
		Content: []mcp.Content{
			&mcp.TextContent{
				Text: "Workspace has no state versions to list.",
			},
		},
	}
}
```

Marin2409: Yeah, I thought of that. The thing is, you’d have to return something like:

```go
return result, &StateVersionSummaryList{Items: summaries, ...}, nil
```

When `result` is non-nil alongside a typed `Out`, the SDK combines both into the response, so the model would see the text message and an empty structured object at the same time.

I’d probably keep it simple and just return the empty list with a `nil` error and no extra content. Semantically, `Items: []` is already pretty unambiguous and its less noise.

Unless we rather to keep the same behavior as the HCPT tests for now, and then we can define a consistent response pattern for `list tools` as part of the migration?

glbyktjys: The extra message indeed repeats what` items: []` already conveys. My intention was to give the agent an explicit explanation of the empty result, but I agree that keeping it simple makes sense. For now, I’ll return the empty list with a nil error, since having no state versions isn’t a tool failure, and update the tests accordingly. We can define a consistent response pattern across the list tools as part of the migration.

PR #521, opened 2026-09-09
[Cleanup tools/toolsets] Fix toolsets default mismatch and clean up naming

Shweta-Deshpande:
## Changes

1. **Bug fix:** **invalid toolset names in "valid" list**
CleanToolsets used to return invalid entries in both the valid list and the invalid list (e.g. passing --toolsets=invalid would list invalid as both "enabled" and "rejected"). Now invalid names only show up in the invalid list, and are correctly excluded from what actually gets enabled.

2. **Replaced hardcoded "all" with the existing toolsets.All constant**
The --toolsets flag's default value was the raw string "all". Swapped it for the existing constant — no behavior change.

3. **Renamed a couple of functions**
- GetValidToolsetNames → ValidToolsetNames
- GetAllValidToolNames → KnownToolNames

4. **Added a new ToolFilter type (pkg/toolsets/filter.go)**
Nothing existing calls it yet. It's a replacement for the old approach of encoding "individual tool mode" string inside a []string of toolset names. ToolFilter{Mode, Toolsets, Tools} makes the two modes (toolset-based vs. individual-tool-based filtering) explicit.
The old sentinel-based EnableIndividualTools / IsToolEnabled functions are left untouched and still used by main.go — they'll be migrated to ToolFilter in a follow-up PR that also refactors dynamic_tool.go, so that migration happens in one place instead of being half-done here.

5. **Tests added**
- TestToolFilter_IndividualModeIgnoresToolsets — cover the new ToolFilter type directly
- TestDefaultKeywordMatchesFlagDefault — locks in the fix from item 2, so the flag default and the default keyword can't silently drift apart again.
- Existing CleanToolsets test cases updated to reflect the bug fix in item 1.

## PCI review checklist

### DIFF
--- a/cmd/terraform-mcp-server/main.go
+++ b/cmd/terraform-mcp-server/main.go
@@ -228,7 +228,7 @@ func parseIndividualTools(toolsFlag string, logger *log.Logger) []string {
 
 	if len(validTools) == 0 {
 		logger.Warn("No valid tools specified, falling back to default toolsets")
-		return parseToolsets("default", logger)
+		return parseToolsets(toolsets.Default, logger)
 	}
 
 	// Use the public API to enable individual tools mode
@@ -266,7 +266,7 @@ func getToolsetsFromCmd(cmd *cobra.Command, logger *log.Logger) []string {
 		toolsetsFlag, err = cmd.Root().PersistentFlags().GetString("toolsets")
 		if err != nil {
 			logger.Warnf("Failed to get toolsets flag, using default: %v", err)
-			toolsetsFlag = "default"
+			toolsetsFlag = toolsets.Default
 		}
 	}
 	return parseToolsets(toolsetsFlag, logger)
--- a/cmd/terraform-mcp-server/main.go
+++ b/cmd/terraform-mcp-server/main.go
@@ -228,7 +228,7 @@ func parseIndividualTools(toolsFlag string, logger *log.Logger) []string {
 
 	if len(validTools) == 0 {
 		logger.Warn("No valid tools specified, falling back to default toolsets")
-		return parseToolsets(toolsets.All, logger)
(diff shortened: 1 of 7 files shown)


cmd/terraform-mcp-server/main.go
```diff
@@ -228,7 +228,7 @@ func parseIndividualTools(toolsFlag string, logger *log.Logger) []string {
 
 	if len(validTools) == 0 {
 		logger.Warn("No valid tools specified, falling back to default toolsets")
-		return parseToolsets("default", logger)
+		return parseToolsets(toolsets.All, logger)
```
jaylonmcshan19-x: Nitpick: 
the fallback in main.go went from` parseToolsets("default")` to `parseToolsets(toolsets.All)`. those aren't the same right? `DefaultToolsets` returns just registry and `All ` will turn everything on including the TFE tools. so something like a typo in --tools used to drop you to registry only and now it enables everything. the log line right above it still says "falling back to default toolsets" too.
Im wondering if that was intentional? there's another `"default"` swap further down in that file that doesn't change any behavior, so wondering if this one got caught up in the same pass.

jaylonmcshan19-x (dismissed):

cmd/terraform-mcp-server/main.go
Shweta-Deshpande: Good catch, you're right, that's a bug — not intentional. Only the flag-default swap in init.go (toolsets.Default → toolsets.All) was needed for hcpt, since that workflow starts the server with no --toolsets flag at all. The parseIndividualTools fallback change is unrelated — no test exercises that branch, which is how it slipped through. Fixing it back to toolsets.Default and adding a test for that fallback path in a follow-up commit.

jaylonmcshan19-x (approved):

PR #522, opened 2026-09-11
[Cleanup tools/toolsets] Replace duplicated tool registration logic with a single tool registry

Shweta-Deshpande:
## What changed
**Cleans up how tools get registered with the MCP server**
Before this change, the same list of ~60 tools was written out by hand, three separate times, each as a long chain of `if this tool is enabled { register it }` blocks (in the mark3labs registry tools, mark3labs TFE tools, and official-SDK tools). There was also a second, hand-maintained map of "which tool belongs to which toolset" that had to be kept in sync by hand with the main list.

Now there is one list of every tool (`pkg/toolsets/registry.go`) that says, per tool: which toolset it's in, whether it needs a TFE session, and whether it needs `ENABLE_TF_OPERATIONS=true`. All three registration spots just loop over this one list instead of repeating themselves. A small map per SDK (`pkg/tools/factories.go`, `pkg/mcp official/tools/tools.go`) says how to actually build each tool.

This also finishes wiring up `ToolFilter` (added in the previous PR) all the way through the app, replacing the old `[]string` + hidden "marker" used to say "individual tool mode."

## What did NOT change
- No tools were added, removed, or renamed.
- No toolset names or defaults changed.
- `ENABLE_TF_OPERATIONS` still gates the exact same 5 tools
  (`delete_team`, `delete_project`, `delete_workspace_safely`,
  `force_unlock_workspace`, `action_run`), and `create_run` still defaults
  to the safe, no-auto-apply version unless that flag is on.

## Notes
- `go.mod`/`go.sum`: bumped, `google.golang.org/grpc` to fix failing build

## PCI review checklist

### DIFF
--- a/cmd/terraform-mcp-server/init.go
+++ b/cmd/terraform-mcp-server/init.go
@@ -250,8 +250,8 @@ func initLogger(outPath string, level log.Level, format string) (*log.Logger, er
 }
 
 // registerToolsAndResources registers tools and resources with the MCP server
-func registerToolsAndResources(hcServer *server.MCPServer, logger *log.Logger, enabledToolsets []string) {
-	tools.RegisterTools(hcServer, logger, enabledToolsets)
+func registerToolsAndResources(hcServer *server.MCPServer, logger *log.Logger, filter toolsets.ToolFilter) {
+	tools.RegisterTools(hcServer, logger, filter)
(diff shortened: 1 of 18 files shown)


jaylonmcshan19-x (approved):
Thanks for the clean up!

PR #525, opened 2026-09-14
rename enableTfOperations to preventDestructiveOperations

jaylonmcshan19-x:
## PCI review checklist

This is the chart side of the `ENABLE_TF_OPERATIONS` rename for [TFECO-13361](https://hashicorp.atlassian.net/browse/TFECO-13361). This PR swaps the env var for `PREVENT_DESTRUCTIVE_OPERATIONS` and the value for `preventDestructiveOperations`, defaults to true so destructive tools are now blocked without any extra config.

one thing I believe its worth noting, I ran into a small gotcha while doing this, the default filter treats false as empty. so if I'd
kept the` | default true` pattern a user setting this to false would've gotten true back and wouldn't have known. So I dropped the filter and let the `values.yaml` supply the default instead. I made sure to test both and true and false both render correctly

This PR depends on [TFECO-13360](https://hashicorp.atlassian.net/browse/TFECO-13360) landing, the env var doesn't do anything until then.

[TFECO-13361]: https://hashicorp.atlassian.net/browse/TFECO-13361?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[TFECO-13360]: https://hashicorp.atlassian.net/browse/TFECO-13360?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Marin2409 (commented):
one question just out of curiosity, is there a reason why you set `preventDestructiveOperations` to `false` in `.github/workflows/helm-lint.yml` and then to `true` in `helm/terraform-mcp-server/values.yaml` ?

jaylonmcshan19-x:
@Marin2409 yeah that's intentional. values.yaml holds the real default which is true, so destructive tools are blocked unless someone opts in. the lint workflow does two template runs, one with defaults, and one with a set of overrides to hit the paths that don't render by default. setting it to false there is just proving the override works. felt worth doing on this one specifically because of the helm default filter thing, if I'd left that in then false would've come back as
true and the override run really wouldn't have caught it.

PR #526, opened 2026-09-16
[Cleanup tools/toolsets] Remove remaining duplicated loops and the last hardcoded switch statement

Shweta-Deshpande:
## What this does

Finishes the tools/toolsets registration cleanup. Ticket: TFECO-13152

Two things were left over after the last PR:

1. Three files (mark3labs registry tools, mark3labs TFE tools, official-SDK
   tools) each had their own copy of the same "loop over every tool, skip
   the disabled ones" code. Replaced with one shared helper,
   `ToolFilter.EnabledTools`.
2. Two tools (`create_no_code_workspace`, `create_run`) were registered
   through a hardcoded `switch` statement instead of a lookup map, like
   every other tool. Moved them into a small `specialFactories` map so
   every tool's setup is data-driven, with no exceptions left in code.

## Testing

- `go build ./...` and `go vet ./...` pass.
- Full test suite passes.
- The factory/registry sync test: `create_no_code_workspace` and `create_run` used to be exempted by name, now they're checked the same way as every other tool.
- Added a test for the new `EnabledTools` helper.

## PCI review checklist


pkg/mcp-official/tools/tools.go
```diff
@@ -27,10 +25,7 @@ var officialFactories = map[string]func(svr *mcp.Server){
 }
 
 func RegisterTools(svr *mcp.Server, logger *slog.Logger, filter toolsets.ToolFilter) {
-	for _, td := range toolsets.AllTools {
-		if !filter.IsToolEnabled(td.Name) {
-			continue
-		}
+	for _, td := range filter.EnabledTools(nil) {
```
jaylonmcshan19-x, line 28: Quick heads up, jose's [PR](https://github.com/hashicorp/terraform-mcp-server/pull/509/changes#diff-f1d2fcab6f03eba4f1e49ba0f45e369ce6809e2b469969a38abdc95de7848169R61) adds a `td.RequiresTFOps` check to this  for delete_project. whichever
of these lands second is gonna need to hold onto that, Just giving ya a heads up so its not lost in the rebase since the func changes shape here.

jaylonmcshan19-x (approved):

PR #532, opened 2026-09-19
Fix list_runs status filter and workspace_name

### LINKED ISSUE #531: list_runs ignores the status filter and returns an empty workspace_name
#### terraform-mcp-server version
```plaintext
1.3.0 (hashicorp/terraform-mcp-server:1.3.0), still present on main at e248187
```

## Description

`list_runs` returns runs, but has two problems:

1. **The `status` filter is ignored.** The tool declares `status` as an array (`mcp.WithArray("status", ...)`), but the handler reads it with `request.GetString("status", "")`. An array is not a string, so `GetString` returns the empty default, `options.Status` is never set, and runs of every status come back.
2. **`workspace_name` is always empty.** The handler copies `r.Workspace.Name`, but neither `RunListOptions` nor `RunListForOrganizationOptions` asks for `include=workspace`, so each run's workspace is a relationship stub with only an ID. This affects both the workspace-scoped and the organization-wide listing.

## Testing plan
```plaintext
list_runs {"terraform_org_name": "<org>", "workspace_name": "<ws>", "status": ["errored"]}
list_runs {"terraform_org_name": "<org>", "status": ["errored", "applied"]}
```

#### Expected Behavior
Only runs with the requested statuses are returned, and each run has its `workspace_name` set.

#### Actual Behavior
Runs of every status are returned (e.g. `applied` and `planned_and_finished` runs for a `["errored"]` filter), and `workspace_name` is `""` on every run.

#### Additional Context
Both fixes are small: read `status` with `GetStringSlice` and join it with commas (the API's `filter[status]` takes a comma-separated list), and set `Include: []tfe.RunIncludeOpt{tfe.RunWorkspace}` on both option structs. A PR with both follows.

pdeva:
Fixes #531.

## What and why

`list_runs` had two bugs, both in `pkg/tools/tfe/list_runs.go`:

- **`status` filter ignored.** The tool declares `status` as an array, but the handler read it with `GetString`, which returns the default for a non-string value. The filter was never sent. It is now read with `GetStringSlice` and joined into the comma-separated list that `filter[status]` expects.
- **`workspace_name` always empty.** Runs were listed without `include=workspace`, so each run's workspace was a relationship stub with only an ID. Both the workspace and organization listings now set `Include: []tfe.RunIncludeOpt{tfe.RunWorkspace}`.

## Testing

Ran the server built from `main` and from this branch over stdio against an HCP Terraform organization:

| Call | `main` | This branch |
|---|---|---|
| workspace + `status: ["errored"]` | `applied`, `errored`, `errored`, `applied`, `errored`; `workspace_name` all `""` | 5 × `errored`, `workspace_name` set |
| org + `status: ["errored","applied"]` | includes `plan_queued`, `planned_and_finished`; `workspace_name` all `""` | only `applied` / `errored`, `workspace_name` set |

`go test ./pkg/tools/...` passes. The existing tests in this package don't exercise handlers against a TFE client, so there is no new unit test.

## PCI review checklist

- [x] If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request. (Reverting the PR is sufficient.)
- [x] If applicable, I've documented the impact of any changes to security controls. (None.)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jaylonmcshan19-x (dismissed):
Thank you for the fix!!

jrhouston (approved):

PR #535, opened 2026-09-21
  Migrate public Registry `provider` tools to the official MCP SDK (1/3)

glbyktjys:
This is the first of three PRs migrating the public Terraform Registry tools to the official MCP SDK. This PR migrates the following provider-related tools:
- `search_providers`
- `get_provider_details`
- `get_latest_provider_version`
- `get_provider_capabilities`

The migrated handlers currently reuse the legacy Registry HTTP helpers, which depend on Logrus. The official MCP server creates an `slog.Logger` during HTTP initialization, but it is not currently injected into tool handlers, so these handlers temporarily use `logrus.StandardLogger()`.

Would love to have any feedback on the preferred logging approach: inject `*slog.Logger` through tool registration and handler, or create official-SDK-specific Registry client helpers whose signatures accept `*slog.Logger`.
  
## PCI review checklist


Shweta-Deshpande:
I suggest going with inject *slog.Logger through registration/handler, not separate SDK-specific client helpers.

`RegisterTools` in the `pkg/mcp-official` directory already receives a real *slog.Logger `server.go line:73` — it just isn't passed to the officialFactories closures. Threading it through (same pattern the mark3labs tools already uses with *logrus.Logger, and that ToolLogging/org_allowlist already use with *slog.Logger) fixes this with no new pattern.

The other option duplicates pkg/client's HTTP/retry/pagination logic, which is shared with the legacy mark3labs tools — which requires two implementations to keep in sync across all 3 migration PRs. So I would not suggest going down that route.

glbyktjys:
> I suggest going with inject *slog.Logger through registration/handler, not separate SDK-specific client helpers.
> 
> `RegisterTools` in the `pkg/mcp-official` directory already receives a real *slog.Logger `server.go line:73` — it just isn't passed to the officialFactories closures. Threading it through (same pattern the mark3labs tools already uses with *logrus.Logger, and that ToolLogging/org_allowlist already use with *slog.Logger) fixes this with no new pattern.
> 
> The other option duplicates pkg/client's HTTP/retry/pagination logic, which is shared with the legacy mark3labs tools — which requires two implementations to keep in sync across all 3 migration PRs. So I would not suggest going down that route.

Got it, thank you! For this migration, I’ll pass the existing `*slog.Logger` through registration into the tool handlers. Since the shared client still requires `*logrus.Logger`, I’ll add a minimal adapter in `pkg/mcp-official/client` that forwards its logs to the injected slog logger. This keeps the existing client signatures and shared HTTP/retry/pagination logic unchanged, and lets PRs 2 and 3 reuse the adapter.

